PR #1, opened 2015-06-30
Override Header Middleware

mikaelm12:
@davidfowl @Tratcher @muratg @victorhurdugaci


test/Microsoft.AspNet.BasicMiddleware.Tests/project.json
```diff
@@ -0,0 +1,28 @@
+﻿{
+  "version": "1.0.0-*",
+  "description": "Microsoft.AspNet.BasicMiddleware.Tests Class Library",
+  "authors": [ "t-mimen" ],
```
davidfowl: Fix the metadata

test/Microsoft.AspNet.BasicMiddleware.Tests/OverrideHeaderMIddlewareTest.cs
```diff
@@ -0,0 +1,208 @@
...
+
+            var req = new HttpRequestMessage(HttpMethod.Get, "");
+            req.Headers.Add("X-Forwarded-For", "11.111.111.11");
+            req.Headers.Add("X-Forwarded-Host", "otherhost");
+            req.Headers.Add("X-Forwarded-Proto", "Protocol");
+            req.Headers.Add("X-Http-Method-Override", "Put");
+            HttpResponseMessage resp = await server.CreateClient().SendAsync(req);
+        }
+
+        public async Task OverrideHeaderMiddleware_DisabledAndEnabled()
```
davidfowl: Missing the Fact attribute

test/Microsoft.AspNet.BasicMiddleware.Tests/OverrideHeaderMIddlewareTest.cs
```diff
@@ -0,0 +1,208 @@
...
+                   return Task.FromResult(0);
+                });
+            });
+
+            var req = new HttpRequestMessage(HttpMethod.Get, "");
+            req.Headers.Add("X-Forwarded-For", "11.111.111.11");
+            req.Headers.Add("X-Forwarded-Host", "otherhost");
+            req.Headers.Add("X-Forwarded-Proto", "Protocol");
+            req.Headers.Add("X-Http-Method-Override", "Put");
+            HttpResponseMessage resp = await server.CreateClient().SendAsync(req);
```
davidfowl: What's the point of this call?

test/Microsoft.AspNet.BasicMiddleware.Tests/OverrideHeaderMIddlewareTest.cs
```diff
@@ -0,0 +1,208 @@
...
+                app.Run(context =>
+                {
+                    Assert.Equal("GET", context.Request.Method);
+                    return Task.FromResult(0);
+                });
+            });
+
+            var req = new HttpRequestMessage(HttpMethod.Get, "X-Forwarded-For Enabled");
+            req.Headers.Add("X-Http-Method-Override", "Put");
+            HttpResponseMessage resp = await server.CreateClient().SendAsync(req);
```
davidfowl: Same here

test/Microsoft.AspNet.BasicMiddleware.Tests/OverrideHeaderMIddlewareTest.cs
```diff
@@ -0,0 +1,208 @@
...
+                    return Task.FromResult(0);
+                });
+            });
+
+            var req = new HttpRequestMessage(HttpMethod.Get, "All header flags true test");
+            req.Headers.Add("X-Forwarded-For", "11.111.111.11");
+            req.Headers.Add("X-Forwarded-Host", "testHost");
+            req.Headers.Add("X-Forwarded-Proto", "Protocol");
+            req.Headers.Add("X-Http-Method-Override", "Put");
+            HttpResponseMessage resp = await server.CreateClient().SendAsync(req);
```
davidfowl: Same here

test/Microsoft.AspNet.BasicMiddleware.Tests/OverrideHeaderMIddlewareTest.cs
```diff
@@ -0,0 +1,208 @@
...
```
davidfowl: ??

src/Microsoft.AspNet.BasicMiddleware/project.json
```diff
@@ -0,0 +1,17 @@
+﻿{
+  "version": "1.0.0-*",
+  "description": "BasicMiddleware Class Library",
+
+  "dependencies": {
+    "System.Collections": "4.0.10-*",
```
davidfowl: Move System.* into dnxcore50 section

src/Microsoft.AspNet.BasicMiddleware/project.json
```diff
@@ -0,0 +1,17 @@
...
+    "System.Linq": "4.0.0-*",
+    "System.Threading": "4.0.10-*",
+    "Microsoft.CSharp": "4.0.0-*",
```
davidfowl: And this as well.

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddlewareOptions.cs
```diff
@@ -0,0 +1,21 @@
...
+            XForwardedForEnabled = allOverridesEnabled;
+            XForwardedHostEnabled = allOverridesEnabled;
+            XForwardedProtoEnabled = allOverridesEnabled;
+            XHttpMethodOverrideEnabled = allOverridesEnabled;
+        }
+
+        public bool XForwardedForEnabled { get; set; }
+        public bool XForwardedHostEnabled { get; set; }
+        public bool XForwardedProtoEnabled { get; set; }
+        public bool XHttpMethodOverrideEnabled { get; set; }
```
davidfowl: Make this a separate middlware.

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddlewareOptions.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.BasicMiddleware
+{
+    public class OverrideHeaderMiddlewareOptions
+    {
+        public OverrideHeaderMiddlewareOptions(bool allOverridesEnabled = true)
```
davidfowl: Make 2 ctors and remove the default parameter value.

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddleware.cs
```diff
@@ -0,0 +1,74 @@
...
+    public class HeaderOverrideMiddleware
+    {
+        private OverrideHeaderMiddlewareOptions _options;
+        private RequestDelegate _next;
+        private const string XForwardedFor = "X-Forwarded-For";
+        private const string XForwardedHost = "X-Forwarded-Host";
+        private const string XForwardedProto = "X-Forwarded-Proto";
+        private const string XHttpMethodOverride = "X-Http-Method-Override";
+
+        public HeaderOverrideMiddleware(RequestDelegate next, OverrideHeaderMiddlewareOptions overrideMiddlewareOptions)
```
davidfowl: Call it options

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddleware.cs
```diff
@@ -0,0 +1,74 @@
...
+        ///     - X-Forward-For
+        ///     - X-Forwarded-Host
+        ///     - X-Forwarded-Proto
+        ///     - X-Http-Method-Override
+        /// </summary>
+        /// <param name="context"></param>
+        /// <returns></returns>
+        public Task Invoke(HttpContext context)
+        {
+            var xForward = context.Request.Headers.GetCommaSeparatedValues(XForwardedFor);
```
davidfowl: Why do you get a list here? Can you put the example format?

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderExtensions.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.BasicMiddleware;
+
+namespace Microsoft.AspNet.Builder
+{
+    public static class HeaderOverrideExtensions
```
Tratcher: The class name doesn't match the file name.

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddleware.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Net;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Builder;
+using Microsoft.AspNet.Http;
+
+namespace Microsoft.AspNet.BasicMiddleware
+{
+    public class HeaderOverrideMiddleware
+    {
+        private OverrideHeaderMiddlewareOptions _options;
```
Tratcher: readonly

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddleware.cs
```diff
@@ -0,0 +1,74 @@
...
+        private RequestDelegate _next;
+        private const string XForwardedFor = "X-Forwarded-For";
```
Tratcher: Put const's at the top.

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddleware.cs
```diff
@@ -0,0 +1,74 @@
...
+        private const string XForwardedHost = "X-Forwarded-Host";
+        private const string XForwardedProto = "X-Forwarded-Proto";
+        private const string XHttpMethodOverride = "X-Http-Method-Override";
+
+        public HeaderOverrideMiddleware(RequestDelegate next, OverrideHeaderMiddlewareOptions overrideMiddlewareOptions)
+        {
+            if (overrideMiddlewareOptions == null || next == null)
```
Tratcher: Remove this check and just use the [NotNull] attribute.

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddleware.cs
```diff
@@ -0,0 +1,74 @@
...
+            if (_options.XForwardedForEnabled && xForward != null && xForward.Count > 0)
+            {
+                IPAddress originalIPAddress;
+                if (IPAddress.TryParse(xForward[0], out originalIPAddress))
+                {
+                    context.Connection.RemoteIpAddress = originalIPAddress;
+                }
+            }
+
+            var xForwardHost = context.Request.Headers.Get(XForwardedHost);
```
davidfowl: Move these inside of the ifs

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddleware.cs
Tratcher: See if you can find a clean pattern that doesn't even access the given request header if not Enabled. Parsing request headers is work we'd like to avoid where possible.

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddleware.cs
```diff
@@ -0,0 +1,74 @@
...
+            if (_options.XForwardedHostEnabled && xForwardHost != null)
+            {
+                context.Request.Host = new HostString(xForwardHost);
```
Tratcher: Set Headers["Host"] directly. The HostString constructor requires a different data format.

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddlewareOptions.cs
Tratcher: @davidfowl why?

test/Microsoft.AspNet.BasicMiddleware.Tests/OverrideHeaderMIddlewareTest.cs
```diff
@@ -0,0 +1,208 @@
...
+                app.UseOverrideHeaders(options);
+                app.Run(context =>
+                {
+                    Assert.Equal("11.111.111.11", context.Connection.RemoteIpAddress.ToString());
+                    return Task.FromResult(0);
+                });
+            });
+
+            var req = new HttpRequestMessage(HttpMethod.Get, "XForwardedFor enabled");
+            req.Headers.Add("X-Forwarded-For", "11.111.111.11");
```
Tratcher: Add tests for multiple values and bad values.

test/Microsoft.AspNet.BasicMiddleware.Tests/OverrideHeaderMIddlewareTest.cs
```diff
@@ -0,0 +1,208 @@
...
+            {
+                app.UseOverrideHeaders(options);
+                app.Run(context =>
+                {
+                    Assert.Equal("11.111.111.11", context.Connection.RemoteIpAddress.ToString());
+                    return Task.FromResult(0);
+                });
+            });
+
+            var req = new HttpRequestMessage(HttpMethod.Get, "XForwardedFor enabled");
```
Tratcher: Why are you setting the path field?

test/Microsoft.AspNet.BasicMiddleware.Tests/OverrideHeaderMIddlewareTest.cs
```diff
@@ -0,0 +1,208 @@
...
+            var req = new HttpRequestMessage(HttpMethod.Get, "All header flags true test");
+            req.Headers.Add("X-Forwarded-For", "11.111.111.11");
+            req.Headers.Add("X-Forwarded-Host", "testHost");
+            req.Headers.Add("X-Forwarded-Proto", "Protocol");
+            req.Headers.Add("X-Http-Method-Override", "Put");
+            HttpResponseMessage resp = await server.CreateClient().SendAsync(req);
+        }
+
+        [Fact]
+        public async Task OverrideHeaderMiddleware_NoOverridesEnabled()
```
Tratcher: This test looks like a duplicate of the one after it.

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        public static IApplicationBuilder UseOverrideHeaders(this IApplicationBuilder builder)
+        {
+            var options = new OverrideHeaderMiddlewareOptions();
+            return UseOverrideHeaders(builder, options);
+        }
```
BrennanConroy: newline

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddlewareOptions.cs
davidfowl: It just is. The other things are headers, this is a form value. They all over override in the name but this kinda looks different IMO. Also other frameworks allow other advanced features in the method override like being able to set the method via an arbitrary callback

Tratcher: The current implementation fully covers this scenario: http://www.hanselman.com/blog/HTTPPUTOrDELETENotAllowedUseXHTTPMethodOverrideForYourRESTServiceWithASPNETWebAPI.aspx
That seems appropriate for this middleware.

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddleware.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNet.Builder;
+using Microsoft.AspNet.Http;
+
+namespace Microsoft.AspNet.BasicMiddleware
+{
+    public class HeaderOverrideMiddleware
+    {
+        private OverrideHeaderMiddlewareOptions _options;
+        private RequestDelegate _next;
```
victorhurdugaci: Readonly here too

test/Microsoft.AspNet.BasicMiddleware.Tests/project.json
```diff
@@ -0,0 +1,28 @@
...
+  "tags": [ "" ],
+  "projectUrl": "",
+  "licenseUrl": "",
+
+  "dependencies": {
+    "System.Collections": "4.0.10-*",
+    "System.Linq": "4.0.0-beta-22913",
+    "System.Threading": "4.0.10-*",
+    "Microsoft.CSharp": "4.0.0-*",
```
victorhurdugaci: Move to dnxcore50

test/Microsoft.AspNet.BasicMiddleware.Tests/project.json
```diff
@@ -0,0 +1,28 @@
...
+    "Microsoft.AspNet.StaticFiles": "1.0.0-*",
+    "Microsoft.AspNet.TestHost": "1.0.0-*",
+    "Microsoft.Framework.Logging.Testing": "1.0.0-*",
+    "xunit.runner.aspnet": "2.0.0-aspnet-*",
```
victorhurdugaci: sort

test/Microsoft.AspNet.BasicMiddleware.Tests/OverrideHeaderMIddlewareTest.cs
```diff
@@ -0,0 +1,208 @@
...
+            var options = new OverrideHeaderMiddlewareOptions(false);
+            TestServer server = TestServer.Create(app =>
+            {
+                app.UseOverrideHeaders(options);
+                app.Run(context =>  {
+                    Assert.Null(context.Connection.RemoteIpAddress);
+                    Assert.Equal("localhost", context.Request.Host.ToString());
+                    Assert.Equal("HTTP/1.1", context.Request.Protocol);
+                    Assert.Equal("GET",context.Request.Method);
+                   return Task.FromResult(0);
```
victorhurdugaci: Format this file. The `[Fact]` attribute is also one off

build.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
```
troydai: Can't check it here in the PR. But please make sure it is executable.
`chmod +x`

test/Microsoft.AspNet.BasicMiddleware.Tests/OverrideHeaderMIddlewareTest.cs
```diff
@@ -0,0 +1,208 @@
...
+            var req = new HttpRequestMessage(HttpMethod.Get, "XForwardedFor enabled");
+            req.Headers.Add("X-Forwarded-For", "11.111.111.11");
+            HttpResponseMessage resp = await server.CreateClient().SendAsync(req);
+        }
+
+        [Fact]
+        public async Task OverrideHeaderMiddleware_XForwardedHostEnabled()
+        {
+            var options = new OverrideHeaderMiddlewareOptions();
+            TestServer server = TestServer.Create(app =>
```
troydai: Generally we use `var` here instead.

test/Microsoft.AspNet.BasicMiddleware.Tests/OverrideHeaderMIddlewareTest.cs
```diff
@@ -0,0 +1,208 @@
...
+                app.Run(context =>
+                {
+                    Assert.Equal("testHost", context.Request.Host.ToString());
+                    return Task.FromResult(0);
+                });
+            });
+
+            var req = new HttpRequestMessage(HttpMethod.Get, "X-Forwarded-For Enabled");
+            req.Headers.Add("X-Forwarded-Host", "testHost");
+            HttpResponseMessage resp = await server.CreateClient().SendAsync(req);
```
troydai: Use `var` as well. And multiple cases in other places.

troydai: And in a case, some how a bug causes the line of code on _43_ is not executed, how to capture this issue?

troydai: And a lot of similar test logic. Can these code be abstracted?

build.sh
dougbu: Use `#!/usr/bin/env bash` shebang instead. Should not have this one in any aspnet repo.

Tratcher: @dougbu Then you should update the template in Universe.

dougbu: @Tratcher the template on `dev` branch was updated back on the 14th.

Tratcher: Ah, maybe he pulled this from master.

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddleware.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// </summary>
+        /// <param name="context"></param>
+        /// <returns></returns>
+        public Task Invoke(HttpContext context)
+        {
+            var xForward = context.Request.Headers.GetCommaSeparatedValues(XForwardedFor);
+            if (_options.XForwardedForEnabled && xForward != null && xForward.Count > 0)
+            {
+                IPAddress originalIPAddress;
+                if (IPAddress.TryParse(xForward[0], out originalIPAddress))
```
davidfowl: We had a discussion yesterday about the security of XFF. There are some things to consider here that we might want to make harder for end users to mess up.

halter73: Is the XFF header the only security concern?

What if the client tries to spoof the protocol/host/etc... using some of the other X-* headers? Or are we relying on the proxy to stomp the relevant X-* headers since this middleware will be opt-in.

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddlewareOptions.cs
```diff
@@ -0,0 +1,21 @@
...
+        {
+            XForwardedForEnabled = allOverridesEnabled;
+            XForwardedHostEnabled = allOverridesEnabled;
+            XForwardedProtoEnabled = allOverridesEnabled;
+            XHttpMethodOverrideEnabled = allOverridesEnabled;
+        }
+
+        public bool XForwardedForEnabled { get; set; }
```
halter73: These four boolean properties could be replaced by a [single enum](https://msdn.microsoft.com/en-us/library/system.flagsattribute(v=vs.110).aspx).

This enum type could be used as the ctor param too for greater flexibility.

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddleware.cs
```diff
@@ -0,0 +1,76 @@
...
+namespace Microsoft.AspNet.BasicMiddleware
+{
+    public class OverrideHeaderMiddleware
+    {
+        private const string XForwardedFor = "X-Forwarded-For";
+        private const string XForwardedHost = "X-Forwarded-Host";
+        private const string XForwardedProto = "X-Forwarded-Proto";
+        private readonly OverrideHeaderMiddlewareOptions _options;
+        private readonly RequestDelegate _next;
+
```
BrennanConroy: remove space

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddleware.cs
```diff
@@ -0,0 +1,76 @@
...
+            if (options == null || next == null)
+            {
+                throw new ArgumentNullException(nameof(options));
+            }
+            _options = options;
+            _next = next;
+        }
+
+        /// <summary>
+        ///  Read the values for the different header keys we are looking for
```
BrennanConroy: Remove a space before Read

test/Microsoft.AspNet.BasicMiddleware.Tests/HttpMethodOverrideMiddleware.cs
```diff
@@ -0,0 +1,50 @@
...
+using Microsoft.AspNet.Builder;
+using Microsoft.AspNet.TestHost;
+using Xunit;
+
+namespace Microsoft.AspNet.BasicMiddleware.Tests
+{
+    public class HttpMethodOverrideMiddlewareTest
+    {
+        [Fact]
+        public async Task HttpMethodOverride_HeaderAvailable()
```
BrennanConroy: Add TestsExecuted to this test

BrennanConroy: Use consistent naming for this variable for all tests. Also, lowercase first word.

test/Microsoft.AspNet.BasicMiddleware.Tests/project.json
```diff
@@ -0,0 +1,27 @@
+﻿{
+  "version": "1.0.0-*",
+  "description": "Microsoft.AspNet.BasicMiddleware.Tests Class Library",
+  "dependencies": {
+    "Microsoft.AspNet.BasicMiddleware": "1.0.0-*",
+    "Microsoft.AspNet.Http.Abstractions": "1.0.0-beta6-11657",
```
BrennanConroy: Don't have specific versions

test/Microsoft.AspNet.BasicMiddleware.Tests/project.json
```diff
@@ -0,0 +1,27 @@
...
+  "commands": {
+    "test": "xunit.runner.aspnet"
+  },
+
+  "frameworks": {
+    "dnx451": { },
+    "dnxcore50": {
+      "dependencies": {
+        "System.Collections": "4.0.10-*",
+        "System.Linq": "4.0.0-beta-22913",
```
BrennanConroy: Don't have specific versions

BrennanConroy:
Add the CI files *.yml

src/Microsoft.AspNet.BasicMiddleware/HttpMethodOverrideMiddleware.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Builder;
+using Microsoft.AspNet.Http;
+
+namespace Microsoft.AspNet.BasicMiddleware
+{
+    // You may need to install the Microsoft.AspNet.Http.Abstractions package into your project
```
victorhurdugaci: huh?

src/Microsoft.AspNet.BasicMiddleware/HttpMethodOverrideMiddleware.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
```
victorhurdugaci: Copyright comment

src/Microsoft.AspNet.BasicMiddleware/HttpMethodOverrideMiddleware.cs
```diff
@@ -0,0 +1,40 @@
...
+            if(xHttpMethodOverrideValue != null)
+            {
+                context.Request.Method = xHttpMethodOverrideValue;
+            }
+            return _next(context);
+        }
+    }
+
+    // Extension method used to add the middleware to the HTTP request pipeline.
+    public static class HttpMethodOverrideExtensions
```
victorhurdugaci: 1. Don't add extension methods in a public nested class
2. Use Xml comments for documentation
3. The comment is wrong. It says "Extension method" but that's a class

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddleware.cs
```diff
@@ -0,0 +1,76 @@
...
+        private const string XForwardedProto = "X-Forwarded-Proto";
+        private readonly OverrideHeaderMiddlewareOptions _options;
+        private readonly RequestDelegate _next;
+
+
+        public OverrideHeaderMiddleware(RequestDelegate next, OverrideHeaderMiddlewareOptions options)
+        {
+            if (options == null || next == null)
+            {
+                throw new ArgumentNullException(nameof(options));
```
victorhurdugaci: Separate in two conditions. Throw with `nameof(next)` if next is null, otherwise it is confusing.

src/Microsoft.AspNet.BasicMiddleware/HttpMethodOverrideMiddleware.cs
```diff
@@ -0,0 +1,40 @@
...
+{
+    // You may need to install the Microsoft.AspNet.Http.Abstractions package into your project
+    public class HttpMethodOverrideMiddleware
+    {
+        private const string XHttpMethodOverride = "X-Http-Method-Override";
+        private readonly RequestDelegate _next;
+
+        public HttpMethodOverrideMiddleware(RequestDelegate next)
+        {
+            _next = next;
```
victorhurdugaci: Null check?

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddleware.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// <returns></returns>
+        public Task Invoke(HttpContext context)
+        {
+            if (_options.XForwardedForEnabled)
+            {
+                var xForward = context.Request.Headers.GetCommaSeparatedValues(XForwardedFor);
+                if (xForward != null && xForward.Count > 0)
+                {
+                    IPAddress originalIPAddress;
+                    if (IPAddress.TryParse(xForward[0], out originalIPAddress))
```
victorhurdugaci: Should you do something (throw?) if `TryParse` fails?

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddlewareOptions.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.BasicMiddleware
+{
+    public class OverrideHeaderMiddlewareOptions
+    {
+        public OverrideHeaderMiddlewareOptions()
+        {
+            XForwardedForEnabled = true;
```
victorhurdugaci: You can call a constructor from another one. In you case, you can just call the other constructor:

```
public OverrideHeaderMiddlewareOptions()
    :this(allOverridesEnabled: true) {
}
```

test/Microsoft.AspNet.BasicMiddleware.Tests/OverrideHeaderMIddlewareTest.cs
```diff
@@ -0,0 +1,185 @@
...
+using Xunit;
+
+namespace Microsoft.AspNet.BasicMiddleware
+{
+    public class OverrideMiddlewareHeaderTests
+    {
+        [Fact]
+        public async Task OverrideHeaderMiddleware_XForwardEnabled()
+        {
+            var AssertsExecuted = false;
```
victorhurdugaci: lowerCaseFirstLetter

test/Microsoft.AspNet.BasicMiddleware.Tests/HttpMethodOverrideMiddleware.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System;
```
victorhurdugaci: Copyright

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddlewareOptions.cs
```diff
@@ -0,0 +1,25 @@
...
+            XForwardedHostEnabled = true;
+            XForwardedProtoEnabled = true;
+        }
+        public OverrideHeaderMiddlewareOptions(bool allOverridesEnabled)
```
victorhurdugaci: space before ctor

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddleware.cs
victorhurdugaci: Is the expected behavior that you forward the first ip address? What if there is more than one?

test/Microsoft.AspNet.BasicMiddleware.Tests/OverrideHeaderMIddlewareTest.cs
```diff
@@ -0,0 +1,185 @@
...
+            req.Headers.Add("X-Forwarded-Proto", "Protocol");
+            await server.CreateClient().SendAsync(req);
+            Assert.True(AssertsExecuted);
+        }
+
+        [Fact]
+        public async Task OverrideHeaderMiddleware_DisabledAndEnabled()
+        {
+            var AssertsExecuted = false;
+            var options = new OverrideHeaderMiddlewareOptions(false);
```
victorhurdugaci: Nit: you can initialize properties inline:

```
var options = new OverrideHeaderMiddlewareOptions(false)
{
    XForwardedForEnabled = true,
    XForwardedProtoEnabled = true
};

File: test/Microsoft.AspNet.BasicMiddleware.Tests/HttpMethodOverrideMiddleware.cs [hunk 48]
```diff
@@ -0,0 +1,50 @@
...
+            });
+            var req = new HttpRequestMessage(HttpMethod.Get, "");
+            req.Headers.Add("X-Http-Method-Override", "Put");
+            await server.CreateClient().SendAsync(req);
+        }
+
+        [Fact]
+        public async Task HttpMethodOverride_HeaderNotAvailable()
+        {
+            var TestsExecuted = false;
```
victorhurdugaci: lowerCaseFirstLetter

test/Microsoft.AspNet.BasicMiddleware.Tests/HttpMethodOverrideMiddleware.cs
```diff
@@ -0,0 +1,50 @@
...
+            var server = TestServer.Create(app =>
+            {
+                app.UseHttpMethodOverrideMiddleware();
+                app.Run(context =>
+                {
+                    Assert.Null(context.Request.Headers.Get("Put"));
```
victorhurdugaci: Shouldn't the argument be `X-Http-Method-Override` instead of `Put`?

test/Microsoft.AspNet.BasicMiddleware.Tests/HttpMethodOverrideMiddleware.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        [Fact]
+        public async Task HttpMethodOverride_HeaderAvailable()
+        {
+            var server = TestServer.Create(app =>
+            {
+                app.UseHttpMethodOverrideMiddleware();
+                app.Run(context =>
+                {
+                    Assert.Equal("Put", context.Request.Headers.Get("X-Http-Method-Override"));
```
victorhurdugaci: 1. You forgot to check if this code is actually executed before ending the test
2. You should check against `context.Request.Method` not the header because that will always be set since you did that a few lines below.

test/Microsoft.AspNet.BasicMiddleware.Tests/OverrideHeaderMIddlewareTest.cs
```diff
@@ -0,0 +1,185 @@
...
+
+using System.Net.Http;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Builder;
+using Microsoft.AspNet.TestHost;
+using Xunit;
+
+namespace Microsoft.AspNet.BasicMiddleware
+{
+    public class OverrideMiddlewareHeaderTests
```
victorhurdugaci: file name != class name

test/Microsoft.AspNet.BasicMiddleware.Tests/HttpMethodOverrideMiddleware.cs
victorhurdugaci: Actually, just like above, you should check the `Request.Method` property...

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddleware.cs
Tratcher: Don't throw. At most you'd log.

.vs/config/applicationhost.config
```diff
@@ -0,0 +1,1030 @@
+﻿<?xml version="1.0" encoding="UTF-8"?>
```
Eilon: Don't check this file in.

src/Microsoft.AspNet.BasicMiddleware/HttpMethodOverrideExtensions.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.Builder;
+
+namespace Microsoft.AspNet.BasicMiddleware
```
Eilon: This class should go in the Builder namespace. For example, look at the extension methods for all the other middlesware: https://github.com/aspnet/Mvc/blob/faaba481e82f15e83b6ff02241fe2c55e568a9d0/src/Microsoft.AspNet.Mvc.Core/BuilderExtensions.cs#L11

src/Microsoft.AspNet.BasicMiddleware/HttpMethodOverrideMiddleware.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        private const string xHttpMethodOverride = "X-Http-Method-Override";
+        private readonly RequestDelegate _next;
+        private readonly string _inputName;
+
+        public HttpMethodOverrideMiddleware(RequestDelegate next, string inputName = null)
+        {
+            _next = next;
+            _inputName = inputName;
+
```
Eilon: Lots of extra newlines in this file.

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddleware.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            if (next == null)
+            {
+                throw new ArgumentNullException(nameof(_next));
+            }
+            _options = options;
+            _next = next;
+        }
+
+        /// <summary>
```
Eilon: The `Invoke` method doesn't really need doc comments because users never call this method. On the other hand, lots of other types/members in this PR need doc comments (e.g. the `IApplicationBuilder` extension methods).

src/Microsoft.AspNet.BasicMiddleware/HttpMethodOverrideMiddleware.cs
```diff
@@ -0,0 +1,44 @@
...
+            }
+
+            var xHttpMethodOverrideValue = context.Request.Headers.Get(xHttpMethodOverride);
+            if (xHttpMethodOverrideValue != null)
+            {
+                context.Request.Method = xHttpMethodOverrideValue;
+            }
+
+            await _next(context);
+
```
victorhurdugaci: empty line

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddlewareOptions.cs
```diff
@@ -0,0 +1,21 @@
...
+using System;
+
+namespace Microsoft.AspNet.BasicMiddleware
+{
+    public class OverrideHeaderMiddlewareOptions
+    {
+        public OverrideHeaderMiddlewareOptions() : this(ForwardedHeaders.All)
+        {
+        }
+        public OverrideHeaderMiddlewareOptions(ForwardedHeaders forwardedOptions)
```
Eilon: Add newline between the ctors.

src/Microsoft.AspNet.BasicMiddleware/project.json
```diff
@@ -0,0 +1,18 @@
+﻿{
+  "version": "1.0.0-*",
+  "description": "BasicMiddleware Class Library",
```
Eilon: Need a real description for this.

src/Microsoft.AspNet.BasicMiddleware/HttpMethodOverrideMiddleware.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        public async Task Invoke(HttpContext context)
+        {
+            if (_inputName != null)
+            {
+                var form = await context.Request.ReadFormAsync();
+                var methodType = form[_inputName];
+                context.Request.Method = methodType;
+                await _next(context);
```
victorhurdugaci: make some changes to not duplicate the await and the assignment to Request.method

src/Microsoft.AspNet.BasicMiddleware/OverrideHeaderMiddlewareOptions.cs
```diff
@@ -0,0 +1,21 @@
...
+        {
+            ForwardedOptions = forwardedOptions;
+        }
+
+        public ForwardedHeaders ForwardedOptions { get; }
```
victorhurdugaci: Probably you can pass the enum directly instead of this class

test/Microsoft.AspNet.BasicMiddleware.Tests/OverrideHeaderMIddlewareTest.cs
```diff
@@ -0,0 +1,206 @@
...
+                    return Task.FromResult(0);
+                });
+            });
+
+            var req = new HttpRequestMessage(HttpMethod.Get, "");
+            req.Headers.Add("X-Forwarded-For", "11.111.111.11");
+            await server.CreateClient().SendAsync(req);
+            Assert.True(assertsExecuted);
+        }
+        [Fact]
```
Eilon: Add newline

test/Microsoft.AspNet.BasicMiddleware.Tests/OverrideHeaderMIddlewareTest.cs
```diff
@@ -0,0 +1,206 @@
...
+using Microsoft.AspNet.Builder;
+using Microsoft.AspNet.TestHost;
+using Xunit;
+
+namespace Microsoft.AspNet.BasicMiddleware
+{
+    public class OverrideMiddlewareHeaderTests
+    {
+        [Fact]
+        public async Task OverrideHeaderMiddleware_XForwardEnabled()
```
Eilon: Please read the [engineering guidelines](https://github.com/aspnet/Home/wiki/Engineering-guidelines#unit-test-method-naming) on good unit test naming patterns.

test/Microsoft.AspNet.BasicMiddleware.Tests/project.json
```diff
@@ -0,0 +1,27 @@
+﻿{
+  "version": "1.0.0-*",
+  "description": "Microsoft.AspNet.BasicMiddleware.Tests Class Library",
```
Eilon: Doesn't need a description.

src/Microsoft.AspNet.BasicMiddleware/ForwardedHeaders.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    [Flags]
+    public enum ForwardedHeaders
+    {
+        None = 0,
+        XForwardedFor = 1,
+        XForwardedHost = 2,
+        XForwardedProto = 4,
+        All = XForwardedFor | XForwardedHost | XForwardedProto
+    }
```
victorhurdugaci: Suggestion:
 
```
None = 0,  
XForwardedFor = 1 << 0,  
XForwardedHost = 1 << 1,  
XForwardedProto = 1 << 2,  
All = XForwardedFor | XForwardedHost | XForwardedProto  
```

davidfowl:
The middleware should not be in a project called BasicMiddleware. Each middleware component group should be in a project with the right name:

Microsoft.AspNet.HttpOverrides or something similar

src/Microsoft.AspNet.HttpOverrides/HttpMethodOverrideExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+namespace Microsoft.AspNet.Builder
+{
+    public static class HttpMethodOverrideExtensions
+    {
+        /// <summary>
+        /// Allows the user to override the default method type of a request
+        /// </summary>
+        /// <param name="builder"></param>
+        /// <returns></returns>
+        public static IApplicationBuilder UseHttpMethodOverrideMiddleware(this IApplicationBuilder builder)
```
Eilon: These methods names probably don't need the `Middleware` suffix.

src/Microsoft.AspNet.HttpOverrides/HttpMethodOverrideExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.HttpOverrides;
+
+namespace Microsoft.AspNet.Builder
+{
+    public static class HttpMethodOverrideExtensions
+    {
+        /// <summary>
+        /// Allows the user to override the default method type of a request
```
Eilon: End sentences with a period.

Also, does this allow the "user" to override the method, or does it allow a "web request" to override it?

src/Microsoft.AspNet.HttpOverrides/HttpMethodOverrideExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+        public static IApplicationBuilder UseHttpMethodOverrideMiddleware(this IApplicationBuilder builder)
+        {
+            return builder.Use(next => new HttpMethodOverrideMiddleware(next).Invoke);
+        }
+
+        /// <summary>
+        /// Allows the user to override the default method type of a request
+        /// </summary>
+        /// <param name="builder"></param>
+        /// <param name="inputName">Denotes the element that contains the name of the resulting method type</param>
```
Eilon: This parameter seems like it's most specifically `formFieldName`, is that the case?

src/Microsoft.AspNet.HttpOverrides/OverrideHeaderMiddleware.cs
```diff
@@ -0,0 +1,71 @@
...
+            {
+                throw new ArgumentNullException(nameof(_next));
+            }
+            _options = options;
+            _next = next;
+        }
+
+        public Task Invoke(HttpContext context)
+        {
+            if ((_options.ForwardedOptions & ForwardedHeaders.XForwardedFor) != 0)
```
blowdart: A couple of things here worry me - 1) We're throwing away the original RemoteIpAddress, which we should just map into another header and 2) are we really considering scenarios where there are multiple different forwarding headers? Because we're making up a precedence order if this happens. You know people will just use All, because they're lazy.

src/Microsoft.AspNet.HttpOverrides/HttpMethodOverrideMiddleware.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            if (_inputName != null)
+            {
+                var form = await context.Request.ReadFormAsync();
+                var methodType = form[_inputName];
+                context.Request.Method = methodType;
+            }
+            else
+            {
+                var xHttpMethodOverrideValue = context.Request.Headers.Get(xHttpMethodOverride);
```
Eilon: My recollection of this is that this is supposed to override only when the current *actual* method is a POST. At least, that's what we did in MVC and/or Web API before. I believe there are security implications to allowing a GET method to change itself to being a DELETE, etc.

@blowdart

src/Microsoft.AspNet.HttpOverrides/OverrideHeaderExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.HttpOverrides;
+
+namespace Microsoft.AspNet.Builder
+{
+    public static class OverrideHeaderExtensions
+    {
+        /// <summary>
+        /// Uses X-Forwarded-For, X-Forwarded-Host, X-Forwarded-Proto, override headers.
```
Eilon: Might want to put some text here is will help a user understand what this is for.

src/Microsoft.AspNet.HttpOverrides/HttpMethodOverrideMiddleware.cs
Eilon: Actually maybe not so bad because you can't add headers on an innocent GET request... but it might still make sense to lock this down a bit.

src/Microsoft.AspNet.HttpOverrides/project.json
```diff
@@ -0,0 +1,18 @@
+﻿{
+  "version": "1.0.0-*",
+  "description": "Various middleware for ASP.NET 5",
```
Eilon: It's not "various" anymore - we should list what's in here.

Also, I'm not sure if we should say the "ASP.NET 5" vs. just "ASP.NET". @danroth27 guidance?

src/Microsoft.AspNet.HttpOverrides/OverrideHeaderMiddleware.cs
```diff
@@ -0,0 +1,71 @@
...
+                if (xForwardHost != null)
+                {
+                    context.Request.Host = new HostString(xForwardHost);
+                }
+            }
+
+            if ((_options.ForwardedOptions & ForwardedHeaders.XForwardedProto) != 0)
+            {
+                var xForwardProto = context.Request.Headers.Get(XForwardedProto);
+                if (XForwardedProto != null)
```
PinpointTownes: You mixed up the constant name and the variable name. Use `xForwardProto` here or you'll get a nasty surprise :smile:

src/Microsoft.AspNet.HttpOverrides/OverrideHeaderMiddleware.cs
```diff
@@ -0,0 +1,71 @@
...
+                {
+                    context.Request.Protocol = xForwardProto;
```
PinpointTownes: Paf, you went directly to the sand trap :scream: 

`HttpRequest.Protocol` (like `OwinContext.Protocol` btw) returns the HTTP protocol version, but it doesn't indicate whether the HTTP request was encapsulated in a SSL connection or not.

You should rather use `HttpRequest.Scheme` here.

src/Microsoft.AspNet.HttpOverrides/HttpMethodOverrideMiddleware.cs
blowdart: You're assuming innocence of requests @Eilon which is not good. But a quick search shows this sort ogf thing being limited to POSTs in other frameworks (Django, ExpressJS)

Also X-Http-Method is another header which does the same thing. (Darned OData)

src/Microsoft.AspNet.HttpOverrides/HttpMethodOverrideExtensions.cs
mikaelm12: Yeah. Thats a much better way to put it

src/Microsoft.AspNet.HttpOverrides/OverrideHeaderMiddleware.cs
mikaelm12: Good catch, thanks!

src/Microsoft.AspNet.HttpOverrides/OverrideHeaderExtensions.cs
blowdart: NO OXFORD COMMAS

Eilon: We *do* use Oxford commas, like it or not. However, the issue here is that that last comma is incorrect either way because the list has already ended.

src/Microsoft.AspNet.HttpOverrides/OverrideHeaderMiddleware.cs
```diff
@@ -0,0 +1,71 @@
...
+        public OverrideHeaderMiddleware(RequestDelegate next, OverrideHeaderMiddlewareOptions options)
+        {
+            if (options == null)
+            {
+                throw new ArgumentNullException(nameof(options));
+            }
+
+            if (next == null)
+            {
+                throw new ArgumentNullException(nameof(_next));
```
BrennanConroy: next not _next

Eilon: Better yet, use `[NotNull]` and avoid all this.

src/Microsoft.AspNet.HttpOverrides/HttpMethodOverrideMiddleware.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        public HttpMethodOverrideMiddleware(RequestDelegate next, string formFieldName = null)
+        {
+            _next = next;
+            _formFieldName = formFieldName;
+        }
+
+        public async Task Invoke(HttpContext context)
+        {
+            if (context.Request.Method == "POST")
```
Tratcher: string.Equals OrdinalIgnoreCase

src/Microsoft.AspNet.HttpOverrides/HttpMethodOverrideMiddleware.cs
```diff
@@ -0,0 +1,47 @@
...
+            {
+                if (_formFieldName != null && context.Request.HasFormContentType)
```
Tratcher: These conditions make for some odd behaviors. Consider the following requests:
- _formFieldName is not set, Form POST, has override header, override header applied
- _formFieldName is set, Form POST without a matching _formFieldName, has override header, override header not applied

test/Microsoft.AspNet.HttpOverrides.Tests/project.json
```diff
@@ -0,0 +1,26 @@
+﻿{
+  "version": "1.0.0-*",
+  "dependencies": {
+    "Microsoft.AspNet.HttpOverrides": "1.0.0-*",
+    "Microsoft.AspNet.Http.Abstractions": "1.0.0-*",
+    "Microsoft.AspNet.StaticFiles": "1.0.0-*",
```
Tratcher: StaticFiles?

test/Microsoft.AspNet.HttpOverrides.Tests/project.json
```diff
@@ -0,0 +1,26 @@
+﻿{
+  "version": "1.0.0-*",
+  "dependencies": {
+    "Microsoft.AspNet.HttpOverrides": "1.0.0-*",
+    "Microsoft.AspNet.Http.Abstractions": "1.0.0-*",
```
Tratcher: You don't need to list any transient dependencies (e.g. anything already referenced by HttpOverrides).

Tratcher:
@blowdart you wanted both of these components moved from `src` to `samples`, correct?

davidfowl:
What? I don't understand, what is moving to be a sample?

blowdart:
If they're not going to be samples then everything needs to be disabled by default.

src/Microsoft.AspNet.HttpOverrides/OverrideHeaderMiddleware.cs
mikaelm12: What header should the IP go into?

blowdart: Make one up :) X-Original-IPAddress? What happens if there are multiple ones though?

Tratcher: This header is already a list, no? Shouldn't you just append to the other end of the list?

mikaelm12: I just added it to the end of the X-Forwarded-For header

src/Microsoft.AspNet.HttpOverrides/OverrideHeaderMiddleware.cs
```diff
@@ -0,0 +1,68 @@
...
+                    }
+                }
+            }
+
+            if ((_options.ForwardedOptions & ForwardedHeaders.XForwardedHost) != 0)
+            {
+                var xForwardHostHeaderValue = context.Request.Headers.Get(XForwardedHostHeaderName);
+                if (!string.IsNullOrEmpty(xForwardHostHeaderValue))
+                {
+                    context.Request.Host = new HostString(xForwardHostHeaderValue);
```
Tratcher: Use HostString.FromUriComponent to ensure the format is correct.

src/Microsoft.AspNet.HttpOverrides/OverrideHeaderMiddlewareOptions.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Microsoft.AspNet.HttpOverrides
+{
+    public class OverrideHeaderMiddlewareOptions
+    {
+        public OverrideHeaderMiddlewareOptions() : this(ForwardedHeaders.None)
+        {
+        }
+
+        public OverrideHeaderMiddlewareOptions(ForwardedHeaders forwardedOptions)
```
Tratcher: Options classes shouldn't have constructor parameters, make the property settable instead.

src/Microsoft.AspNet.HttpOverrides/OverrideHeaderExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+namespace Microsoft.AspNet.Builder
+{
+    public static class OverrideHeaderExtensions
+    {
+        /// <summary>
+        /// Forwards proxied headers onto current request
+        /// </summary>
+        /// <param name="builder"></param>
+        /// <returns></returns>
+        public static IApplicationBuilder UseOverrideHeaders(this IApplicationBuilder builder)
```
Tratcher: Remove this overload, it doesn't do anything because everything is turned off by default.

mikaelm12:
Updated! @Tratcher

Tratcher:
:shipit:

Eilon:
:dancers:

PR #6, opened 2015-07-31
ResponseBuffering middleware

Tratcher:
https://github.com/aspnet/Hosting/issues/299

This PR seeks to formalize the discussion from https://github.com/aspnet/HttpAbstractions/wiki/Rolling-Notes about how response buffering components should behave.  In this case we've implemented a fully buffered response. Buffering is disabled via a feature interface or by calling Flush.

The one notable departure from the notes is that we use Stream.SetLength to truncate/reset the buffer, where Position is just used to move around your write marker.

@davidfowl @lodejard @Eilon

### DIFF
--- a/src/Microsoft.AspNet.Buffering/BufferingWriteStream.cs
+++ b/src/Microsoft.AspNet.Buffering/BufferingWriteStream.cs
@@ -1,0 +1,218 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.Buffering
+{
+    internal class BufferingWriteStream : Stream
+    {
+        private readonly Stream _innerStream;
+        private readonly MemoryStream _buffer = new MemoryStream();
+        private bool _isBuffering = true;
+
+        public BufferingWriteStream(Stream innerStream)
+        {
+            _innerStream = innerStream;
+        }
+
+        public override bool CanRead
+        {
+            get { return false; }
+        }
+
+        public override bool CanSeek
+        {
+            get { return _isBuffering; }
+        }
+
+        public override bool CanWrite
+        {
+            get { return _innerStream.CanWrite; }
+        }
+
+        public override long Length
+        {
+            get
+            {
+                if (_isBuffering)
+                {
+                    return _buffer.Length;
+                }
+                // May throw
+                return _innerStream.Length;
+            }
+        }
+
+        // Clear/Reset the buffer by setting Position, Seek, or SetLength to 0. Random access is not supported.
+        public override long Position
+        {
+            get
+            {
+                if (_isBuffering)
+                {
+                    return _buffer.Position;
+                }
+                // May throw
+                return _innerStream.Position;
+            }
+            set
+            {
+                if (_isBuffering)
+                {
+                    if (value != 0)
+                    {
+                        throw new ArgumentOutOfRangeException(nameof(value), value, nameof(Position) + " can only be set to 0.");
+                    }
+                    _buffer.Position = value;
+                    _buffer.SetLength(value);
+                }
+                else
+                {
+                    // May throw
+                    _innerStream.Position = value;
+                }
+            }
+        }
+
+        // Clear/Reset the buffer by setting Position, Seek, or SetLength to 0. Random access is not supported.
+        public override void SetLength(long value)
+        {
+            if (_isBuffering)
+            {
+                if (value != 0)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(value), value, nameof(Length) + " can only be set to 0.");
+                }
+                _buffer.Position = value;
+                _buffer.SetLength(value);
+            }
+            else
+            {
+                // May throw
+                _innerStream.SetLength(value);
+            }
+        }
+
+        // Clear/Reset the buffer by setting Position, Seek, or SetLength to 0. Random access is not supported.
+        public override long Seek(long offset, SeekOrigin origin)
+        {
+            if (_isBuffering)
+            {
+                if (origin != SeekOrigin.Begin)
+                {
+                    throw new ArgumentException(nameof(origin), nameof(Seek) + " can only be set to " + nameof(SeekOrigin.Begin) + ".");
+                }
+                if (offset != 0)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(offset), offset, nameof(Seek) + " can only be set to 0.");
+                }
+                _buffer.SetLength(offset);
+                return _buffer.Seek(offset, origin);
+            }
+            // Try the inner stream instead, but this will usually fail.
+            return _innerStream.Seek(offset, origin);
+        }
+
+        internal void DisableBuffering()
+        {
+            _isBuffering = false;
+            if (_buffer.Length > 0)
+            {
+                Flush();
+            }
+        }
+
+        internal Task DisableBufferingAsync(CancellationToken cancellationToken)
+        {
+            _isBuffering = false;
+            if (_buffer.Length > 0)
+            {
+                return FlushAsync(cancellationToken);
+            }
+            return Task.FromResult(0);
+        }
+
+        public override void Write(byte[] buffer, int offset, int count)
+        {
+            if (_isBuffering)
+            {
+                _buffer.Write(buffer, offset, count);
+            }
+            else
+            {
+                _innerStream.Write(buffer, offset, count);
+            }
+        }
+
+        public override Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            if (_isBuffering)
+            {
+                return _buffer.WriteAsync(buffer, offset, count, cancellationToken);
+            }
+            else
+            {
+                return _innerStream.WriteAsync(buffer, offset, count, cancellationToken);
+            }
+        }
+#if !DNXCORE50
+        public override IAsyncResult BeginWrite(byte[] buffer, int offset, int count, AsyncCallback callback, object state)
+        {
+            if (_isBuffering)
+            {
+                return _buffer.BeginWrite(buffer, offset, count, callback, state);
+            }
+            else
+            {
+                return _innerStream.BeginWrite(buffer, offset, count, callback, state);
+            }
+        }
+
+        public override void EndWrite(IAsyncResult asyncResult)
+        {
+            if (_isBuffering)
+            {
+                _buffer.EndWrite(asyncResult);
+            }
+            else
+            {
+                _innerStream.EndWrite(asyncResult);
+            }
+        }
+#endif
+        public override void Flush()
+        {
+            _isBuffering = false;
+            if (_buffer.Length > 0)
+            {
+                _buffer.Seek(0, SeekOrigin.Begin);
+                _buffer.CopyTo(_innerStream);
+                _buffer.Seek(0, SeekOrigin.Begin);
+                _buffer.SetLength(0);
+            }
+            _innerStream.Flush();
+        }
+
+        public override async Task FlushAsync(CancellationToken cancellationToken)
+        {
+            _isBuffering = false;
+            if (_buffer.Length > 0)
+            {
+                _buffer.Seek(0, SeekOrigin.Begin);
+                await _buffer.CopyToAsync(_innerStream, 1024 * 16, cancellationToken);
+                _buffer.Seek(0, SeekOrigin.Begin);
+                _buffer.SetLength(0);
+            }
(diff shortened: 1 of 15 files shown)


src/Microsoft.AspNet.Buffering/project.json
```diff
@@ -0,0 +1,16 @@
+﻿{
+  "version": "1.0.0-*",
+  "description": "Microsoft.AspNet.Buffering Class Library",
```
Eilon: Description could use some improvement :sa:

src/Microsoft.AspNet.Buffering/BufferingWriteStream.cs
```diff
@@ -0,0 +1,136 @@
...
+using System;
+using System.IO;
+
+namespace Microsoft.AspNet.Buffering
+{
+    internal class BufferingWriteStream : Stream
+    {
+        private readonly Stream _innerStream;
+        private readonly MemoryStream _buffer = new MemoryStream();
+        private bool _shouldBuffer = true;
```
Eilon: Is there a reason that this uses the term "should"? Is it merely a suggestion, or is it stating factually whether buffering is to be used at this moment? Looking at the code it *seems* to me that it is the latter.

src/Microsoft.AspNet.Buffering/BufferingWriteStream.cs
```diff
@@ -0,0 +1,136 @@
...
+            }
+            // Try the inner stream instead, but this will usually fail.
+            return _innerStream.Seek(offset, origin);
+        }
+
+        public override void Write(byte[] buffer, int offset, int count)
+        {
+            if (_shouldBuffer)
+            {
+                // TODO: buffer size limit?
```
Eilon: Should resolve this TODO comment or log a bug.

src/Microsoft.AspNet.Buffering/ResponseBufferingMiddlewareExtensions.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.Builder;
+
+namespace Microsoft.AspNet.Buffering
+{
+    public static class ResponseBufferingMiddlewareExtensions
+    {
+        public static IApplicationBuilder UseResponseBuffering(this IApplicationBuilder builder)
```
Eilon: Doc comment on this method please (it's very user-facing).

src/Microsoft.AspNet.Buffering/ResponseBufferingMiddleware.cs
```diff
@@ -0,0 +1,55 @@
...
+        private readonly RequestDelegate _next;
+
+        public ResponseBufferingMiddleware(RequestDelegate next)
+        {
+            _next = next;
+        }
+
+        public async Task Invoke(HttpContext httpContext)
+        {
+            // TODO: shim the SendFile func, treat it as an auto-flush
```
Eilon: Resolve or file a bug?

Eilon:
Just some fairly small comments, the code overall looks good. I like the new unit tests.

Tratcher:
Open question: Should there be a (configurable) limit on how much data to buffer before disabling buffering and flushing?

Tratcher:
Updated

src/Microsoft.AspNet.Buffering/BufferingWriteStream.cs
```diff
@@ -0,0 +1,186 @@
...
+
+        public override void EndWrite(IAsyncResult asyncResult)
+        {
+            if (_isBuffering)
+            {
+                base.EndWrite(asyncResult);
+            }
+            else
+            {
+                base.EndWrite(asyncResult);
```
Eilon: _innerStream?

src/Microsoft.AspNet.Buffering/ResponseBufferingMiddlewareExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.Builder;
+
+namespace Microsoft.AspNet.Buffering
+{
+    public static class ResponseBufferingMiddlewareExtensions
+    {
+        /// <summary>
+        /// Enabled full buffering of response bodies. This can be disabled on a per request basis using IHttpBufferingFeature.
```
Eilon: Not sure where my comment went... `Enables` on the first word here.

src/Microsoft.AspNet.Buffering/ResponseBufferingMiddleware.cs
```diff
@@ -0,0 +1,58 @@
...
+                if (originalSendFileFeature != null)
+                {
+                    httpContext.SetFeature<IHttpSendFileFeature>(new SendFileFeatureWrapper(originalSendFileFeature, bufferStream));
+                }
+
+                await _next(httpContext);
+
+                // If we're still buffered, set the content-length header and flush the buffer.
+                // Only if the content-length header is not already set, and some content was buffered.
+                if (!httpContext.Response.HasStarted && bufferStream.CanSeek && bufferStream.Length > 0)
```
kichalla: if nothing has been written should we explicitly set the Content-Length to 0 (if it was not already set)?

src/Microsoft.AspNet.Buffering/SendFileFeatureWrapper.cs
```diff
@@ -0,0 +1,28 @@
...
+        public SendFileFeatureWrapper(IHttpSendFileFeature originalSendFileFeature, BufferingWriteStream bufferStream)
+        {
+            _originalSendFileFeature = originalSendFileFeature;
+            _bufferStream = bufferStream;
+        }
+
+        // Flush and disable the buffer if anyone tries to call the SendFile feature.
+        public Task SendFileAsync(string path, long offset, long? length, CancellationToken cancellation)
+        {
+            _bufferStream.DisableBuffering();
```
kichalla: Nice, I think we have some follow up items on MVC side where we have FileStreamResult, StreamOutputFormatter and we should do similar thing

src/Microsoft.AspNet.Buffering/ResponseBufferingMiddleware.cs
Tratcher: no, let the server figure that out. Content-Length is also not valid with some status codes.

Eilon:
Looks :shipit: to me.

src/Microsoft.AspNet.Buffering/HttpBufferingFeature.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.Http.Features;
+
+namespace Microsoft.AspNet.Buffering
+{
+    internal class HttpBufferingFeature : IHttpBufferingFeature
```
kichalla: Q : Curious what is the idea for exceptions thrown when serializers throw exceptions(example in MVC)? Like if we know the response is being buffered, then we can send back a friendly error message. But with this interface it doesn't seem to be possible to check 'if' the response is buffered...

kichalla:
:shipit:

src/Microsoft.AspNet.Buffering/HttpBufferingFeature.cs
Tratcher: You can check HttpResponse.HasStarted or HttpResponse.Body.CanSeek to know if the response body is buffered and/or if the headers can still be modified. Then Set Position and SetLength 0 zero to clear the body.

That said, MVC probably shouldn't render serialization exceptions inline. Throw it and let the exception handling logic decide.

Eilon: Yeah I'd be surprised if *MVC* does anything.

rynowak: We do something here, but it's easily circumvented. What @Tratcher is suggesting is a better choice than what we have.

rynowak: See aspnet/Mvc#2900

Tratcher:
@davidfowl @lodejard waiting on an explicit sign-off to confirm the design concepts.

src/Microsoft.AspNet.Buffering/ResponseBufferingMiddleware.cs
```diff
@@ -0,0 +1,58 @@
...
+        private readonly RequestDelegate _next;
+
+        public ResponseBufferingMiddleware(RequestDelegate next)
+        {
+            _next = next;
+        }
+
+        public async Task Invoke(HttpContext httpContext)
+        {
+            var originalBufferingFeature = httpContext.GetFeature<IHttpBufferingFeature>();
```
lodejard: if `httpContext.Response.HasStarted`== true, or `originalResponseBody.CanSeek` == true, then this middleware should probably no-op instead of wrap.

Tratcher: If HasStarted, Invoke shouldn't even be called. Do we have any scenarios where middleware writes to the response and then calls next?

src/Microsoft.AspNet.Buffering/HttpBufferingFeature.cs
lodejard: I believe there was also talk of HttpResponse convenience API for that case:

```
bool HttpResponse.CanClear { get{
  true if Response.Body.CanSeek && !ResponseHeadersSent
}}
void HttpResponse.Clear() {
  reset body and headers.
}
```

lodejard: Since they're very straightforward we should add these methods also, can you do that as this PR?

Tratcher: This is the wrong repo/project to add them to. I think they'd go into Microsoft.AspNet.Http.Extensions.

src/Microsoft.AspNet.Buffering/BufferingWriteStream.cs
```diff
@@ -0,0 +1,186 @@
...
+        }
+
+        public override bool CanRead
+        {
+            get { return false; }
+        }
+
+        public override bool CanSeek
+        {
+            get { return _isBuffering || _innerStream.CanSeek; }
```
lodejard: Safe to assume CanSeek is always false at this point, so return only _isBuffering?

src/Microsoft.AspNet.Buffering/BufferingWriteStream.cs
```diff
@@ -0,0 +1,186 @@
...
+        public override bool CanWrite
+        {
+            get { return _innerStream.CanWrite; }
+        }
+
+        public override long Length
+        {
+            get
+            {
+                if (_isBuffering)
```
lodejard: Hate to say it, but it's probably a good idea to maintain a `long _length` in this class that's adjusted when Write's, Seek, set Position, and SendFile are called... That way the behavior of the stream's Position and Length props can be deterministic and non-throwing in all cases...

It should also probably throw an exception all of the time when _isBuffering == false when someone calls Seek or set Position

src/Microsoft.AspNet.Buffering/BufferingWriteStream.cs
```diff
@@ -0,0 +1,186 @@
...
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.Buffering
+{
+    internal class BufferingWriteStream : Stream
+    {
+        private readonly Stream _innerStream;
+        private readonly MemoryStream _buffer = new MemoryStream();
```
lodejard: Might want to create a new issue to adjust this a bit eventually to accommodate arbitrarily large responses. For example, this could be a HybridStream that keeps the first 64k in memory (like in non-linear 8k memory pooled segments), and buffers any data over that size in a temp file (prefereable delete-on-close).

src/Microsoft.AspNet.Buffering/HttpBufferingFeature.cs
lodejard: Ha, d'oh! Oh course.

src/Microsoft.AspNet.Buffering/BufferingWriteStream.cs
Tratcher: https://github.com/aspnet/BasicMiddleware/issues/8

Tratcher:
Updated with Louis's feedback about preventing random access to the buffer. Added a sample.

Tratcher:
@lodejard

PR #7, opened 2015-08-03
Mikaelm12/originalValueHeaders

mikaelm12:
Storing original IP, host, and protocol values in new headers.
@victorhurdugaci  @Tratcher @blowdart

### DIFF
--- a/src/Microsoft.AspNet.HttpOverrides/OverrideHeaderMiddleware.cs
+++ b/src/Microsoft.AspNet.HttpOverrides/OverrideHeaderMiddleware.cs
@@ -15,6 +15,9 @@ namespace Microsoft.AspNet.HttpOverrides
         private const string XForwardedForHeaderName = "X-Forwarded-For";
         private const string XForwardedHostHeaderName = "X-Forwarded-Host";
         private const string XForwardedProtoHeaderName = "X-Forwarded-Proto";
+        private const string XOriginalIPName = "X-Original-IP";
+        private const string XOriginalHostName = "X-Original-Host";
+        private const string XOriginalProtoName = "X-Original-Proto";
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.HttpOverrides/OverrideHeaderMiddleware.cs
```diff
@@ -31,21 +34,27 @@ public Task Invoke(HttpContext context)
                 var xForwardedForHeaderValue = context.Request.Headers.GetCommaSeparatedValues(XForwardedForHeaderName);
                 if (xForwardedForHeaderValue != null && xForwardedForHeaderValue.Count > 0)
                 {
-                    IPAddress originalIPAddress;
-                    if (IPAddress.TryParse(xForwardedForHeaderValue[0], out originalIPAddress))
+                    IPAddress IPFromHeader;
```
Tratcher: nit: lower case variable name ipFromHeader

src/Microsoft.AspNet.HttpOverrides/OverrideHeaderMiddleware.cs
```diff
@@ -31,21 +34,27 @@ public Task Invoke(HttpContext context)
...
                     }
                 }
             }
 
             if ((_options.ForwardedOptions & ForwardedHeaders.XForwardedHost) != 0)
             {
+                var hostString = context.Request.Host.ToString();
+                if (!string.IsNullOrEmpty(hostString))
+                {
+                    context.Request.Headers.Set(XOriginalHostName, hostString);
```
Tratcher: You should only set this if you're going to modify Request.Host. Same for Scheme.

mikaelm12:
Updated

Tratcher:
:shipit:

Issue #19, opened 2015-12-03
OverrideHeaderMiddleware doesn't handle X-Forwarded-For with ports


Tratcher:
Code from IISPlatformHandler that should be moved to OverrideHeaderMiddleware to correctly handle ports.
```
        private static void UpdateRemoteIp(HttpContext httpContext)
        {
            var xForwardedForHeaderValue = httpContext.Request.Headers.GetCommaSeparatedValues(XForwardedForHeaderName);
            if (xForwardedForHeaderValue != null && xForwardedForHeaderValue.Length > 0)
            {
                IPAddress ipFromHeader;
                int? port;
                if (IPAddressWithPortParser.TryParse(xForwardedForHeaderValue[0], out ipFromHeader, out port))
                {
                    var connection = httpContext.Connection;
                    var remoteIPString = connection.RemoteIpAddress?.ToString();
                    if (!string.IsNullOrEmpty(remoteIPString))
                    {
                        httpContext.Request.Headers[XOriginalIPName] = remoteIPString;
                    }
                    if (port.HasValue)
                    {
                        if (connection.RemotePort != 0)
                        {
                            httpContext.Request.Headers[XOriginalPortName] = connection.RemotePort.ToString(CultureInfo.InvariantCulture);
                        }
                        connection.RemotePort = port.Value;
                    }
                    connection.RemoteIpAddress = ipFromHeader;
                }
            }
        }
```
```
// Copyright (c) .NET Foundation. All rights reserved.
// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.

using System.Net;
using Microsoft.AspNet.IISPlatformHandler;
using Xunit;

namespace Microsoft.AspNet.PipelineHandler.Tests
{
    public class IPAddressWithPortParserTests
    {
        [Theory]
        [InlineData("127.0.0.1", "127.0.0.1", null)]
        [InlineData("127.0.0.1:1", "127.0.0.1", 1)]
        [InlineData("1", "0.0.0.1", null)]
        [InlineData("1:1", "0.0.0.1", 1)]
        [InlineData("::1", "::1", null)]
        [InlineData("[::1]", "::1", null)]
        [InlineData("[::1]:1", "::1", 1)]
        public void ParsesCorrectly(string input, string expectedAddress, int? expectedPort)
        {
            IPAddress address;
            int? port;
            var success = IPAddressWithPortParser.TryParse(input, out address, out port);
            Assert.True(success);
[...]

Issue #2, opened 2015-07-09
HttpMethodOverrideMiddleware needs Options class


mikaelm12:
Needs to be opt in

PR #20, opened 2015-12-15
#2 add options, #19 fix IP parser, update extensions.

Tratcher:
#2, #19, @muratg @blowdart @pakrym 
Clean up the extensions, add an options class.
Bring over the address parser from IISPlatformHandler.
Store the original IP and port in X-Original-For.
Add a sample.

### DIFF
--- a/samples/HttpOverridesSample/Startup.cs
+++ b/samples/HttpOverridesSample/Startup.cs
@@ -1,0 +1,36 @@
+﻿using Microsoft.AspNet.Builder;
+using Microsoft.AspNet.Hosting;
+using Microsoft.AspNet.Http;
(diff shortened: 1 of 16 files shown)


samples/HttpOverridesSample/Startup.cs
```diff
@@ -0,0 +1,36 @@
+﻿using Microsoft.AspNet.Builder;
```
muratg: License

Tratcher: Not in samples?

muratg: :ok:

pakrym:
:ship: :it:

PR #24, opened 2016-01-11
Added Unit Tests for form field based HTTP Method Override

tthiery:
I was just thinking about prototyping the issue I filed, when I came across that there are some unit tests missing in the field of HTTP Method Override based on form data.


victorhurdugaci:
@Tratcher can you please review this?

PR #30, opened 2016-02-17
Enabled xml doc generation

ajaybhargavb:
@NTaylorMullen / @pranavkm


src/Microsoft.AspNetCore.HttpOverrides/HttpMethodOverrideExtensions.cs
```diff
@@ -28,8 +27,7 @@ public static IApplicationBuilder UseHttpMethodOverride(this IApplicationBuilder
         /// Allows incoming POST request to override method type with type specified in form.
         /// </summary>
         /// <param name="builder"></param>
```
NTaylorMullen: Since you filled in the other you should fill this in too.

Something like `The <see cref="IApplicationBuilder"/> instance this method extends.`

Same below.

NTaylorMullen:
:shipit:

Issue #34, opened 2016-03-11
Add HTTP compression middleware


DamianEdwards:
Add middleware for doing HTTP response compression. Should support GZIP and Brotli. Would need to be configurable to allow control of:
- enabled compression algorithms (maybe support compression providers?)
- response content types to compress (should default to a known good list)
- buffering behavior

E.g.:

``` c#
public void Configure(IApplicationBuilder app)
{
    app.UseResponseCompression();
    app.UseStaticFiles();
    app.Run(async context =>
    {
        context.Response.ContentType = "text/plain";
        await context.Response.WriteAsync("Hello compressed!");
    });
}
```

AlirezaP:
good idea.
in addition its not bad  for add some attribute like "https://github.com/AlirezaP/TagCompressor.Net" and "https://github.com/AlirezaP/TagCompressor.NetCore".

AlirezaP:
@DamianEdwards what's your opinion about this?
[AP.AspNet.ResponseCompression](https://github.com/AlirezaP/AP.AspNet.ResponseCompression/tree/master)
it's a middleware for ResponseCompression but it's not for GZIP or Brotli.

DamianEdwards:
You're talking about HTML minification there, rather than compression. It's also interesting but IMO is much better handled at compile time rather than run time. This issue is specifically about adding middleware to support [HTTP compression](https://en.wikipedia.org/wiki/HTTP_compression).

AlirezaP:
i'm get it.
thank you.

AlirezaP:
@DamianEdwards  
I have added gzip and it worked, but first request.
I mean in the first request, the response page is null (context.Response.Body) but when you refresh the page (just once) it works correctly after that.
[AP.AspNet.ResponseCompression ](https://github.com/AlirezaP/AP.AspNet.ResponseCompression/blob/master/src/AP.AspNet.ResponseCompression/ResponseGzipCompressMiddleware.cs)
this is in runtime, isn't it?
if yes, would you please tell me how I can handle it in compile time?

Tratcher:
@AlirezaP Your Invoke method is still very broken which is why it doesn't work for the first request.  Let's discuss this in your repo instead.

AlirezaP:
sure. thanks

AndreiGorlov:
@AlirezaP 
> I have added gzip and it worked, but first request.
[...]

PR #52, opened 2016-08-04
Response Compression middleware

Yves57:
A first PR proposition about #34 

Included:
- GZIP compression.
- Options to set compression level, content minimum size threshold, accepted MIME types, additional compression providers.

Pending:
- Probably optimizations in the temporary `MemoryStream` management.
- A default accepted MIME types list?
- A default value for `MinimumSize` option?
- Other compression algorithms?


src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionMiddleware.cs
```diff
@@ -0,0 +1,147 @@
...
+            {
+                await _next(context);
+                return;
+            }
+
+            var bodyStream = context.Response.Body;
+
+            using (var uncompressedStream = new MemoryStream())
+            {
+                context.Response.Body = uncompressedStream;
```
Tratcher: Fully buffering all responses is not viable. We'll need to wrap the Response.Body stream and do the compression check on the first write.

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionMiddleware.cs
```diff
@@ -0,0 +1,147 @@
...
+                    }
+                }
+            }
+
+            return null;
+        }
+
+        private bool IsMimeTypeCompressable(string mimeType)
+        {
+            return !string.IsNullOrEmpty(mimeType) && _mimeTypes.Contains(mimeType);
```
Tratcher: What about optional parameters like charset?

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionMiddleware.cs
Yves57: Is there any special test to add, or I can assume that in any cases the MIME type is available when the first byte is written in the body stream?

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionMiddleware.cs
Yves57: Sorry! I will cut the MIME type string to the first ';', and trim the spaces. And I suppose that the string must match the right case.

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionMiddleware.cs
```diff
@@ -86,37 +86,45 @@ public ResponseCompressionMiddleware(RequestDelegate next, IOptions<ResponseComp
...
+
+                                if (compressedStream.Length >= uncompressedStream.Length)
+                                {
+                                    uncompressedStream.Seek(0, SeekOrigin.Begin);
+                                    await uncompressedStream.CopyToAsync(bodyStream);
+                                }
+                                else
+                                {
+                                    context.Response.Headers[HeaderNames.ContentEncoding] = compressionProvider.EncodingName;
+                                    context.Response.Headers[HeaderNames.ContentMD5] = StringValues.Empty;      // Reset the MD5 because the content changed.
```
Tratcher: Headers.Remove...

src/Microsoft.AspNetCore.ResponseCompression/BodyWrapperStream.cs
```diff
@@ -0,0 +1,185 @@
...
+            return _bodyOriginalStream.WriteAsync(buffer, offset, count, cancellationToken);
+        }
+
+        private void OnWrite()
+        {
+            if (!_compressionChecked)
+            {
+                if (IsCompressable())
+                {
+                    UncompressedStream = new MemoryStream();
```
Tratcher: Better, at least now you only buffer responses that are compressible. You can take it one step further though. Instead of making a memory stream here you could directly create the compression stream over the original response stream and update the headers immediately. You'll loose your _minimumSize feature (unless you check the to see if the content-length is already set), and you'll have to use chunked rather than content-length, but that should be OK.

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+namespace Microsoft.AspNetCore.ResponseCompression
+{
+    public static class ResponseCompressionExtensions
+    {
+        /// <summary>
+        /// Allows to compress HTTP Responses.
+        /// </summary>
+        /// <param name="builder">The <see cref="IApplicationBuilder"/> instance this method extends.</param>
+        /// <param name="mimeTypes">The accepted MIME types.</param>
+        public static IApplicationBuilder UseResponseCompression(this IApplicationBuilder builder, IEnumerable<string> mimeTypes)
```
davidfowl: Not a fan of this overload.

test/Microsoft.AspNetCore.ResponseCompression.Tests/project.json
```diff
@@ -0,0 +1,26 @@
...
+    "warningsAsErrors": true
+  },
+    "dependencies": {
+        "dotnet-test-xunit": "2.2.0-*",
+        "Microsoft.AspNetCore.TestHost": "1.1.0-*",
+        "Microsoft.AspNetCore.ResponseCompression": "1.1.0-*",
+        "xunit": "2.2.0-*",
+        "Microsoft.AspNetCore.Http.Abstractions": "1.1.0-*",
+        "Microsoft.Net.Http.Headers": "1.1.0-*"
+    },
```
davidfowl: nit: fix the formatting in this file.

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionExtensions.cs
Yves57: I suppose that most users will let all other parameters as default. Maybe UseResponseCompression(this IApplicationBuilder, params string[] mimeTypes)?
Or you prefer an overload without parameter, and a default MIME type list?

Yves57: I will remove this overload in a first time.

And I was thinking that is won't be easy for a user to add something like "text/*". It will have to add MIME types one by one.
It would be nice to have a `MimeTypeSet` more or less like that:

```
class MimeTypeSet
{
   public void Include(Regex mimeType) { ... }
   public void Include(string mimeType) { ... }
   public void Exclude(Regex mimeType) { ... }
   public void Exclude(string mimeType) { ... }

   public bool Contains(string mimeType) { ... }
}
```
Maybe in the `Microsoft.AspNetCore.WebUtilities` assembly?

src/Microsoft.AspNetCore.ResponseCompression/BodyWrapperStream.cs
```diff
@@ -0,0 +1,201 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
+using System;
```
troydai: Please put System.* ahead of other namespaces.

samples/ResponseCompressionSample/ResponseCompressionSample.xproj
```diff
@@ -0,0 +1,21 @@
...
+    <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
+  </PropertyGroup>
+  <Import Project="$(VSToolsPath)\DNX\Microsoft.DNX.Props" Condition="'$(VSToolsPath)' != ''" />
+  <PropertyGroup Label="Globals">
+    <ProjectGuid>b2a3ce38-51b2-4486-982c-98c380af140e</ProjectGuid>
+    <BaseIntermediateOutputPath Condition="'$(BaseIntermediateOutputPath)'=='' ">.\obj</BaseIntermediateOutputPath>
+    <OutputPath Condition="'$(OutputPath)'=='' ">.\bin\</OutputPath>
+  </PropertyGroup>
+  <PropertyGroup Label="Configuration">
+    <RootNamespace>C:\ws\github\BasicMiddleware\samples\RewriteSample\Properties\AssemblyInfo.csResponseCompressionSample</RootNamespace>
```
troydai: Remove?

src/Microsoft.AspNetCore.ResponseCompression/BodyWrapperStream.cs
```diff
@@ -0,0 +1,201 @@
...
+
+        private readonly HashSet<string> _mimeTypes;
+
+        private readonly IResponseCompressionProvider _compressionProvider;
+
+        private bool _compressionChecked = false;
+
+        private Stream _compressionStream = null;
+
+        internal BodyWrapperStream(HttpResponse response, Stream bodyOriginalStream, HashSet<string> mimeTypes, IResponseCompressionProvider compressionProvider)
```
troydai: Maybe `ISet<string>` instead of `HashSet` if the unique elements are the only concern.

davidfowl:
@Yves57 Can you get some basic performance data when this middleware is active and doing its job.

src/Microsoft.AspNetCore.ResponseCompression/GzipResponseCompressionProvider.cs
```diff
@@ -0,0 +1,31 @@
...
+    public class GzipResponseCompressionProvider : IResponseCompressionProvider
+    {
+        private readonly CompressionLevel _level;
+
+        public GzipResponseCompressionProvider(CompressionLevel level)
+        {
+            _level = level;
+        }
+
+        public string EncodingName
```
troydai: nit: follow form of property can make the code more condense and readable

```
public string EncodingName { get; } = "gzip";
```

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionMiddleware.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// <summary>
+        /// Initialize the Response Compression middleware.
+        /// </summary>
+        /// <param name="next"></param>
+        /// <param name="options"></param>
+        public ResponseCompressionMiddleware(RequestDelegate next, IOptions<ResponseCompressionOptions> options)
+        {
+            if (options.Value.MimeTypes == null)
+            {
+                throw new ArgumentNullException(nameof(options.Value.MimeTypes));
```
troydai: I'm not sure about this exception. `MimeTypes` is not one of the constructor's parameters. This will cause confusion. Instead how about throw `ArgumentException($"{nameof(options.Value.MimeTypes)} is not provided in argument {nameof(options)}")`

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionMiddleware.cs
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        /// <summary>
+        /// Invoke the middleware.
+        /// </summary>
+        /// <param name="context"></param>
+        /// <returns></returns>
+        public async Task Invoke(HttpContext context)
+        {
+            IResponseCompressionProvider compressionProvider = SelectProvider(context.Request.Headers[HeaderNames.AcceptEncoding]);
```
troydai: `var`

blowdart:
I'd like to see a flag, which is set by default, so compression doesn't happen if the request protocol is HTTPS to combat variations on Beast, Crime et al.

Tratcher:
@DamianEdwards @muratg Are we actually planning to accept/ship this as an official package or should this be moved to contrib?

blowdart:
If you're shipping it then I definitely want my flag :D

DamianEdwards:
I'm more than happy to accept this from a function/product point of view (modulo usual API, security, perf, engineering reviews, etc.)

Yves57:
For the HTTPS flag, is `EnableHttps` a correct name for you?

Yves57:
I have used the `TestServer` to do very basic performance tests:
* Body length: 100 bytes (so compression time itself is very small, but of course there is still the compression initialization time)
* 10000 requests
* Test written in a unit test run in Release

The results on my computer:
* No compression middleware: 965 ms
* Compression active but not used: 1025 ms
* Compression active and used : 1500 ms

I will push the perf test source code in my next PR (I can remove it later if you want).

Tratcher:
How does it perform with Kestrel?

Yves57:
@Tratcher @davidfowl 

Other test:
* Using "go-wrk" the client side (with 1 or 4 connections during 10 seconds).
* Using the PR sample application, except that the body length is a 100 byte constant string
* first run result ignored (warmup)

On my computer (so all in local) with 1 connection:
* No compression middleware: 379 requests/s
* Compression active but not used: 379 requests/s
* Compression active and used: 353 requests/s

On my computer (so all in local) with 4 connections:
* No compression middleware: 499 requests/s
* Compression active but not used: 498 requests/s
* Compression active and used: 494 requests/s

So the stream wrapper seems to be more or less "free" in term of performance.

RehanSaeed:
I'm going to be building a Request Compression middleware (I work for a company that sends large JSON logs from Android apps in rural Africa where GPRS internet connectivity is low and compression would make a massive difference). Request Compression is essentially going to be a mirror image of Response Compression.

Is there any appetite for extending this PR or starting a new PR based on this one?

RehanSaeed:
# MIME Types

The MIME types have no default value, I propose the following list which only includes standard MIME types where possible and no common MIME type synonyms like `text/javascript`.

```
var mimeTypes = new string[]
{
    // Plain Text
    "text/html",
    "text/plain",
    "text/css",
    "text/mathml",
    "application/rtf",
    // JSON
    "application/javascript",
    "application/json",
    "application/manifest+json",
    "application/x-web-app-manifest+json",
    "text/cache-manifest",
    // XML
    "application/atom+xml",
    "application/rss+xml",
    "application/xslt+xml",
    "application/xml",
    // Fonts
    "font/opentype",
    "font/otf",
    "font/truetype",
    "application/font-woff",
    "application/vnd.ms-fontobject",
    "application/x-font-ttf",
    // Images
    "image/svg+xml",
    "image/x-icon"
};
```

You'd probably also want to use the above list for Request Compression, although the option to be different should still be there.

# Minimum Size Option

IIS and NGINX provide an options which lets you specify a minimum size before compression is applied. This option should be added to the `ResponseCompressionOptions`.

Yves57:
@davidfowl @Tratcher 
I just made a memory profiling. Here are the results for 3368 requests.

response_compression_memory

I think the most important information in that `DeflateStream` allocates a significant amount of memory (`GZipStream` is in fact a wrapper around `DeflateStream`).
* [A managed buffer of 8192 bytes](https://github.com/dotnet/corefx/blob/master/src/System.IO.Compression/src/System/IO/Compression/DeflateStream.cs#L116) that is allocated each time.
* Unmanaged memory buffers are allocated and destroyed each time in ZLib, while a native function `deflateReset` allows to reset ZLib without destroying buffers.

I think a good optimization would be to add a method `GZipStream.Set(Stream newOutput)` to reset an existing GZipStream without destroying buffers and assign a new output. With this method, we could have `GZipStream` instances in a pool, and use the on demand.

dodyg:
I had to resort using the following MIME type statement instead of just ```application/json``` because that's what  `context.Response.ContentType` returns.

`app.UseResponseCompression(new string[]{ "application/json; charset=utf-8" });         
`

Yves57:
Are you sure that you have the latest version of the PR? The extension method with only MIME types does not exist anymore. And you just tried your sample without any problem.

src/Microsoft.AspNetCore.ResponseCompression/project.json
```diff
@@ -0,0 +1,34 @@
...
+        "repository": {
+            "type": "git",
+            "url": "git://github.com/aspnet/basicmiddleware"
+        },
+        "tags": [
+            "aspnetcore"
+        ]
+    },
+    "dependencies": {
+        "Microsoft.AspNetCore.Http": "1.1.0-*",
```
davidfowl: Middleware shouldn't need to reference Http directly. What are you using from here that isn't in abstractions?

Yves57: You're right, "Http.Abstractions" is enough. I fix it.

dodyg:
@Yves57 ah my fault.  It works now.

```
            var compressionOptions = new ResponseCompressionOptions
            {
                MimeTypes = new string[]{ "application/json" }
            };
            app.UseResponseCompression(compressionOptions);
```

andrewlock:
Just for reference, compression via middleware is available currently in [WebMarkupMin.AspNetCore1](https://github.com/Taritsyn/WebMarkupMin/tree/master/src/WebMarkupMin.AspNetCore1)  - I haven't compared the implementations, but there may be something of interest pertinent to this PR?

src/Microsoft.AspNetCore.ResponseCompression/project.json
```diff
@@ -0,0 +1,34 @@
+{
+    "version": "1.1.0-*",
```
Tratcher: Set this to `0.1.0-*` for the initial checkin. This will let us merge it sooner and iterate on it in the branch until it's ready for release.

src/Microsoft.AspNetCore.ResponseCompression/project.json
```diff
@@ -0,0 +1,34 @@
...
+    "buildOptions": {
+        "warningsAsErrors": true,
+        "keyFile": "../../tools/Key.snk",
+        "nowarn": [
+            "CS1591"
```
Tratcher: Fix this

src/Microsoft.AspNetCore.ResponseCompression/project.json
```diff
@@ -0,0 +1,34 @@
...
+        ],
+        "xmlDoc": true
+    },
+    "description": "ASP.NET Core basic middleware for HTTP Response compression.",
```
Tratcher: remove basic

src/Microsoft.AspNetCore.ResponseCompression/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+// General Information about an assembly is controlled through the following
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyConfiguration("")]
```
Tratcher: Copy from https://github.com/aspnet/BasicMiddleware/blob/dev/src/Microsoft.AspNetCore.HttpOverrides/Properties/AssemblyInfo.cs

test/Microsoft.AspNetCore.ResponseCompression.Tests/ResponseCompressionMiddlewareTest.cs
```diff
@@ -0,0 +1,298 @@
...
+            httpContext.Request.IsHttps = true;
+
+            httpContext.Response.Body = new MemoryStream();
+
+            await middleware.Invoke(httpContext);
+
+            Assert.Equal(expectedLength, httpContext.Response.Body.Length);
+        }
+
+        private async Task<HttpResponse> InvokeMiddleware(int uncompressedBodyLength, string requestAcceptEncoding, string responseType, Action<HttpResponse> addResponseAction = null)
```
Tratcher: Use TestServer to arrange this.  See https://github.com/aspnet/BasicMiddleware/blob/dev/test/Microsoft.AspNetCore.HttpOverrides.Tests/ForwardedHeadersMiddlewareTest.cs#L22

src/Microsoft.AspNetCore.ResponseCompression/BodyWrapperStream.cs
```diff
@@ -0,0 +1,206 @@
...
+            }
+            else
+            {
+                _bodyOriginalStream.Flush();
+            }
+        }
+
+        public override Task FlushAsync(CancellationToken cancellationToken)
+        {
+            if (_compressionStream != null)
```
Tratcher: This needs to call OnWrite() if the user is just trying to flush the headers

src/Microsoft.AspNetCore.ResponseCompression/BodyWrapperStream.cs
```diff
@@ -0,0 +1,206 @@
...
+
+            set
+            {
+                throw new NotSupportedException();
+            }
+        }
+
+        public override void Flush()
+        {
+            if (_compressionStream != null)
```
Tratcher: OnWrite();

samples/ResponseCompressionSample/Properties/launchSettings.json
```diff
@@ -0,0 +1,25 @@
...
+      "applicationUrl": "http://localhost:49658/",
+      "sslPort": 0
+    }
+  },
+  "profiles": {
+    "IIS Express": {
+      "commandName": "IISExpress",
+      "launchBrowser": true,
+      "environmentVariables": {
+        "ASPNET_ENV": "Development"
```
Yves57: I will fix, but just for information "ResponseBufferingSample" has the same error 😄 .

test/Microsoft.AspNetCore.ResponseCompression.Tests/ResponseCompressionMiddlewareTest.cs
Yves57: I have already tried to use `TestServer`, but I cannot neighter check the Content-Encoding header value in `HttpResponseMessage` (it is empty) nor the content length (the decompression is automatically done). Or I miss something?

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionOptions.cs
```diff
@@ -0,0 +1,26 @@
...
+        /// <summary>
+        /// The compression providers. If 'null', the GZIP provider is set as default.
+        /// </summary>
+        public IEnumerable<IResponseCompressionProvider> Providers { get; set; }
+
+        /// <summary>
+        /// 'False' to enable compression only on HTTP request. Enable compression on HTTPS request
+        /// may lead to security problems.
+        /// </summary>
+        public bool EnableHttps { get; set; } = false;
```
blowdart: No it shouldn't. Remove.

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionOptions.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// <summary>
+        /// The compression providers. If 'null', the GZIP provider is set as default.
+        /// </summary>
+        public IEnumerable<IResponseCompressionProvider> Providers { get; set; }
+
+        /// <summary>
+        /// 'False' to enable compression only on HTTP requests. Enable compression on HTTPS requests
+        /// may lead to security problems.
+        /// </summary>
+        public bool EnableHttps { get; set; } = false;
```
Tratcher: We discussed this offline. It can stay, but will remain off by default.  bing, google, msn, and wikipedia all enable this, so we can't outright prohibit it.

Tratcher:
@RehanSaeed there are a few different issues with request compression. A) There's no opportunity for the client and server to negotiate the compression protocol, the client needs to know in advance what the server supports. B) the server may not need to decompress the body for a given operation. E.g. for your compressed log files, what if the web server only wanted to save them to disk in their compressed format. 

Let's not try to put these two behaviors into one middleware. However, there may be some re-usable components between the two.

You can check with @DamianEdwards, but I don't anticipate accepting a contribution for request decompression. https://github.com/aspnet-contrib may be a better place for it.

samples/ResponseCompressionSample/Properties/launchSettings.json
Tratcher: Feel free to fix the ResponseBufferingSample too :-)

Tratcher:
Rebased, squashed, and merged.

PR #63, opened 2016-08-09
Don't use abbreviations

pranavkm:


pranavkm:
cc @ZestyBread

ZestyBread:
:ship: :it:

PR #68, opened 2016-08-18
De-staticifying Parsers

ZestyBread:
Removing static instances for file parsers. General nit-worthy stuff.
@BrennanConroy @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/ConditionPatternParser.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/ConditionPatternParser.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using Microsoft.AspNetCore.Rewrite.Internal;
 
 namespace Microsoft.AspNetCore.Rewrite.Internal.ModRewrite
 {
@@ -10,7 +9,7 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.ModRewrite
     /// Parses the "CondPattern" portion of the RewriteCond. 
     /// RewriteCond TestString CondPattern
     /// </summary>
-    public static class ConditionPatternParser
+    public class ConditionPatternParser
     {
         private const char Not = '!';
         private const char Dash = '-';
@@ -26,7 +25,7 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.ModRewrite
         /// </summary>
         /// <param name="condition">The CondPattern portion of a mod_rewrite RewriteCond.</param>
         /// <returns>A new parsed condition.</returns>
-        public static ParsedModRewriteInput ParseActionCondition(string condition)
+        public ParsedModRewriteInput ParseActionCondition(string condition)
         {
             if (condition ==  null)
             {
@@ -39,7 +38,7 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.ModRewrite
                 throw new FormatException(Resources.FormatError_InputParserUnrecognizedParameter(condition, context.Index));
             }
 
-            // If we hit a !, make sure the condition is inverted when resolving the string
+            // If we hit a !, invert the condition
             if (context.Current == Not)
             {
                 results.Invert = true;
@@ -139,7 +138,7 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.ModRewrite
         /// <param name="context"></param>
         /// <param name="invert"></param>
         /// <returns></returns>
-        public static ParsedModRewriteInput ParseProperty(ParserContext context, bool invert)
+        private static ParsedModRewriteInput ParseProperty(ParserContext context, bool invert)
         {
             if (!context.Next())
             {
--- a/src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/FlagParser.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/FlagParser.cs
@@ -2,22 +2,68 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 
 namespace Microsoft.AspNetCore.Rewrite.Internal.ModRewrite
 {
     public class FlagParser
-    { 
-        public static Flags Parse(string flagString)
+    {
+        private readonly IDictionary<string, FlagType> _ruleFlagLookup = new Dictionary<string, FlagType>(StringComparer.OrdinalIgnoreCase) {
+            { "b", FlagType.EscapeBackreference},
+            { "c", FlagType.Chain },
+            { "chain", FlagType.Chain},
+            { "co", FlagType.Cookie },
+            { "cookie", FlagType.Cookie },
+            { "dpi", FlagType.DiscardPath },
+            { "discardpath", FlagType.DiscardPath },
+            { "e", FlagType.Env},
+            { "env", FlagType.Env},
+            { "end", FlagType.End },
+            { "f", FlagType.Forbidden },
+            { "forbidden", FlagType.Forbidden },
+            { "g", FlagType.Gone },
+            { "gone", FlagType.Gone },
+            { "h", FlagType.Handler },
+            { "handler", FlagType.Handler },
+            { "l", FlagType.Last },
+            { "last", FlagType.Last },
+            { "n", FlagType.Next },
+            { "next", FlagType.Next },
+            { "nc", FlagType.NoCase },
+            { "nocase", FlagType.NoCase },
+            { "ne", FlagType.NoEscape },
+            { "noescape", FlagType.NoEscape },
+            { "ns", FlagType.NoSubReq },
+            { "nosubreq", FlagType.NoSubReq },
+            { "p", FlagType.Proxy },
+            { "proxy", FlagType.Proxy },
+            { "pt", FlagType.PassThrough },
+            { "passthrough", FlagType.PassThrough },
+            { "qsa", FlagType.QSAppend },
+            { "qsappend", FlagType.QSAppend },
+            { "qsd", FlagType.QSDiscard },
+            { "qsdiscard", FlagType.QSDiscard },
+            { "qsl", FlagType.QSLast },
+            { "qslast", FlagType.QSLast },
+            { "r", FlagType.Redirect },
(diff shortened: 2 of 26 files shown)


src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/ConditionPatternParser.cs
```diff
@@ -139,7 +138,7 @@ public static ParsedModRewriteInput ParseActionCondition(string condition)
         /// <param name="context"></param>
         /// <param name="invert"></param>
         /// <returns></returns>
-        public static ParsedModRewriteInput ParseProperty(ParserContext context, bool invert)
+        private ParsedModRewriteInput ParseProperty(ParserContext context, bool invert)
```
BrennanConroy: private doesn't need xml docs, also why private?

Tratcher: no harm if it actually explains something

src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/FlagParser.cs
```diff
@@ -2,22 +2,68 @@
...
+        public Flags Parse(string flagString)
         {
             if (string.IsNullOrEmpty(flagString))
             {
-                return null;
+                throw new ArgumentNullException("flagString is null");
             }
 
             // Check that flags are contained within []
             if (!flagString.StartsWith("[") || !flagString.EndsWith("]"))
```
BrennanConroy: Could you change this to `!(flagString.StartsWith("[") && flagString.EndsWith("]"))`?

src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/FlagParser.cs
```diff
@@ -2,22 +2,68 @@
...
             {
-                throw new FormatException();
+                throw new FormatException("Flags do no start or end with square brackets: [flags]");
```
BrennanConroy: Not a fan of this grammar
"Flags should start and end with square brackets: [flags]"?

src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/ConditionPatternParser.cs
Tratcher: If they're private and don't use any class member data, they can remain static.

src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/FlagParser.cs
```diff
@@ -27,13 +73,20 @@ public static Flags Parse(string flagString)
             foreach (string token in tokens)
             {
                 var hasPayload = token.Split('=');
+
+                FlagType flag;
+                if (!_ruleFlagLookup.TryGetValue(hasPayload[0], out flag))
+                {
+                    throw new FormatException("Unrecognized flag");
```
BrennanConroy: $"Unrecognized flag: {hasPayload[0]}"?

src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/TestStringParser.cs
```diff
@@ -97,7 +96,7 @@ public static Pattern Parse(string testString)
         /// <param name="context">The ParserContext</param>
         /// <param name="results">The List of results which the new condition parameter will be added.</param>
         /// <returns>true </returns>
-        private static void ParseConditionParameter(ParserContext context, List<PatternSegment> results)
+        private void ParseConditionParameter(ParserContext context, List<PatternSegment> results)
```
BrennanConroy: These are also private, so docs not needed?

Tratcher: They're needed for the next guy, probobly you :-)

src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/ConditionPatternParser.cs
ZestyBread: Any Reason to have them be static vs non-static in this case?

src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/TestStringParser.cs
ZestyBread: I wrote them in when I had no idea what I was doing lol. I can delete them for you ;)

src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/ConditionPatternParser.cs
troydai: For a static method you can confidently tell it won't change the state of an instance. It guards the expectation. It also helps to improve readability because the reader can easily make this basic assumption.

src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/ConditionPatternParser.cs
```diff
@@ -217,7 +216,7 @@ public static ParsedModRewriteInput ParseProperty(ParserContext context, bool in
             }
         }
 
-        private static bool IsValidActionCondition(ParsedModRewriteInput results)
+        private bool IsValidActionCondition(ParsedModRewriteInput results)
```
troydai: Same as above. If the function behave the same with and without `static` modifier, leave it on.

src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/FlagParser.cs
```diff
@@ -2,22 +2,68 @@
...
 using System;
+using System.Collections.Generic;
 
 namespace Microsoft.AspNetCore.Rewrite.Internal.ModRewrite
 {
     public class FlagParser
-    { 
-        public static Flags Parse(string flagString)
+    {
+        private IDictionary<string, FlagType> _ruleFlagLookup = new Dictionary<string, FlagType>(StringComparer.OrdinalIgnoreCase) {
```
troydai: Can we make it static and readonly so as to avoid repeatedly allocate this dictionary?

src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/FlagParser.cs
```diff
@@ -2,22 +2,68 @@
...
+            { "t", FlagType.Type },
+            { "type", FlagType.Type },
+        };
+
+        public Flags Parse(string flagString)
         {
             if (string.IsNullOrEmpty(flagString))
             {
-                return null;
+                throw new ArgumentNullException("flagString is null");
```
troydai: `throw new ArgumentNullException(nameof(flagString))`

src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/FlagParser.cs
troydai: Ah, never mind. @Tratcher explains why it is not static.

src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/ConditionPatternParser.cs
ZestyBread: This only runs once on startup, would like to create the instances and have gc clean them up after use.

src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/FlagParser.cs
```diff
@@ -27,13 +73,20 @@ public static Flags Parse(string flagString)
             foreach (string token in tokens)
             {
                 var hasPayload = token.Split('=');
+
+                FlagType flag;
+                if (!_ruleFlagLookup.TryGetValue(hasPayload[0], out flag))
+                {
+                    throw new FormatException("Unrecognized flag: " + hasPayload[0]);
```
troydai: Nit: We begin to use [string interpolation(https://msdn.microsoft.com/en-us/library/dn961160.aspx)] commonly now.

src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/FlagParser.cs
ZestyBread: Still should make it readonly though?

src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/TestStringParser.cs
troydai: Comments are good. Don't delete them. I frequently amazed by my own comments which I left usually not long ago ......

src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/ConditionPatternParser.cs
troydai: This is correct an instance can be collected while a static class's fields won't. But this shouldn't be related to static methods.

src/Microsoft.AspNetCore.Rewrite/Internal/UrlRewrite/FileParser.cs
```diff
@@ -211,14 +212,14 @@ private static void ParseUrlAction(XElement urlAction, UrlRewriteRuleBuilder bui
             }
         }
 
-        private static void ThrowUrlFormatException(XElement element, string message)
+        private void ThrowUrlFormatException(XElement element, string message)
```
Tratcher: statics?

src/Microsoft.AspNetCore.Rewrite/Internal/UrlRewrite/InputParser.cs
```diff
@@ -31,7 +31,7 @@ public static Pattern ParseInputString(string testString)
             return ParseString(context);
         }
 
-        private static Pattern ParseString(ParserContext context)
+        private Pattern ParseString(ParserContext context)
```
Tratcher: statics

BrennanConroy:
Go home :sheep: :it:

PR #69, opened 2016-08-19
Tests for Mod_rewrite

ZestyBread:
More Hardening. 
@Tratcher


test/Microsoft.AspNetCore.Rewrite.Tests/ModRewrite/FlagParserTest.cs
```diff
@@ -47,6 +48,23 @@ public void FlagParser_CheckManyTermsWithEquals()
             Assert.True(DictionaryContentsEqual(results.FlagDictionary, expected.FlagDictionary));
         }
 
+        [Theory]
+        [InlineData("]", "Flags should start and end with square brackets: [flags]")]
+        [InlineData("[", "Flags should start and end with square brackets: [flags]")]
+        [InlineData("[R, L]", "Unrecognized flag:  L")] // cannot have spaces after ,
```
Tratcher: quote the value? It's not obvious the space is the problem

Tratcher:
:shipit:

PR #71, opened 2016-08-22
Fix to tokenizing mod_rewrite rules

ZestyBread:
Escaped characters prior didn't remove the escape character \, which was incorrect. Also, sections can be enclosed by quotation marks ", which are trimmed from rules now.
Notes:

- Currently using .Trim() and .Replace() to do this.
- Characters can be escaped in string enclosed by quotation marks. Hence no matter what we remove the \ characters,

### DIFF
--- a/src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/Tokenizer.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/Tokenizer.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Text.RegularExpressions;
 
 namespace Microsoft.AspNetCore.Rewrite.Internal.ModRewrite
 {
@@ -30,7 +31,7 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.ModRewrite
             }
             var context = new ParserContext(rule);
             context.Next();
-            
+
             var tokens = new List<string>();
             context.Mark();
             while (true)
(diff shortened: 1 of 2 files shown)


ZestyBread:
@Tratcher

src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/Tokenizer.cs
```diff
@@ -74,8 +89,22 @@ public List<string> Tokenize(string rule)
...
+
+        // Need to remove leading and trailing slashes if they exist.
+        // This is on start-up, so more forgivening towards substrings/ new strings
+        // If this is a perf/memory problem, discuss later.
+        private List<string> RemoveQuotesAndEscapeCharacters(List<string> tokens)
+        {
+            var retList = new List<string>();
+            foreach (var token in tokens)
+            {
+                retList.Add(token.Trim('\"').Replace("\\", ""));
```
Tratcher: What if I escape a backslash? E.g. "\\"?

src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/Tokenizer.cs
```diff
@@ -74,8 +90,24 @@ public List<string> Tokenize(string rule)
...
-            
-            return tokens;
+
+            return RemoveQuotesAndEscapeCharacters(tokens);
+        }
+
+        // Need to remove leading and trailing slashes if they exist.
+        // This is on start-up, so more forgivening towards substrings/ new strings
+        // If this is a perf/memory problem, discuss later.
+        private List<string> RemoveQuotesAndEscapeCharacters(List<string> tokens)
```
davidfowl: Why not mutate the incoming list?

ZestyBread: Didn't want to deal with concurrent modification of a list, but can fix.

Tratcher:
:shipit:

PR #75, opened 2016-08-29
Removes RuleResult, added to the context

ZestyBread:
@Tratcher @davidfowl @natemcmaster 
Changes the return value of actions to RuleTermination (an enum) and void to rules.


src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/ModRewriteRedirectAction.cs
```diff
@@ -64,7 +64,7 @@ public override RuleResult ApplyAction(RewriteContext context, MatchResults rule
                     context.HttpContext.Response.Headers[HeaderNames.Location] = pattern + context.HttpContext.Request.QueryString;
                 }
             }
-            return RuleResult.ResponseComplete;
+            return RuleTermination.ResponseComplete;
```
Tratcher: Why not change ApplyAction to void and update the RewriteContext directly?

src/Microsoft.AspNetCore.Rewrite/RewriteMiddleware.cs
```diff
@@ -65,30 +65,31 @@ public Task Invoke(HttpContext context)
...
+                switch (rewriteContext.Results)
                 {
                     case RuleTermination.Continue:
                         _logger.RewriteMiddlewareRequestContinueResults();
                         break;
                     case RuleTermination.ResponseComplete:
                         _logger.RewriteMiddlewareRequestResponseComplete(
-                            urlContext.HttpContext.Response.Headers[HeaderNames.Location],
-                            urlContext.HttpContext.Response.StatusCode);
+                            rewriteContext.HttpContext.Response.Headers[HeaderNames.Location],
```
Tratcher: why not just reference the httpcontext directly?

src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/ModRewriteRedirectAction.cs
ZestyBread: I don't like propagating the updating of the context here. It creates the notion that the action will always update the context, and is required, rather than the rule setting it. Ex RedirectToHttps doesn't have an action and sets the rewrite context directly.

Tratcher: How is that different from requiring the Action to return a RuleTermination?

src/Microsoft.AspNetCore.Rewrite/RewriteContext.cs
```diff
@@ -17,6 +17,7 @@ public class RewriteContext
         public HttpContext HttpContext { get; set; }
         public IFileProvider StaticFileProvider { get; set; }
         public ILogger Logger { get; set; }
+        public RuleTermination Results { get; set; }
```
natemcmaster: Nit: should this be singular? There is only one result

ZestyBread:
:up: :date: I conformed @Tratcher

src/Microsoft.AspNetCore.Rewrite/Internal/ModRewrite/RuleBuilder.cs
```diff
@@ -215,7 +215,7 @@ public void AddRule(string rule)
                 else
                 {
                     var last = flags.HasFlag(FlagType.End) || flags.HasFlag(FlagType.Last);
-                    var redirect = last ? RuleResult.StopRules : RuleResult.Continue;
+                    var redirect = last ? RuleTermination.StopRules : RuleTermination.Continue;
```
Tratcher: why is this variable named redirect?

Tratcher:
:shipit:

troydai:
:shipit:

troydai:
@ZestyBread remember to delete your branch after merge so as to keep the branch list clean.

PR #77, opened 2016-08-29
Use c# back references for regex matching, refactor to rewrite/redirect actions

ZestyBread:
@Tratcher @natemcmaster @davidfowl


ZestyBread:
The diff got enlarged due to file renaming, effectively what I'm doing is using the mod_rewrite implementation with default parameters for UrlRewrite to use.

src/Microsoft.AspNetCore.Rewrite/Internal/UrlActions/RewriteAction.cs
```diff
@@ -38,10 +62,25 @@ public override void ApplyAction(RewriteContext context, MatchResults ruleMatch,
                 FragmentString fragment;
                 UriHelper.FromAbsolute(pattern, out scheme, out host, out path, out query, out fragment);
 
+                if (query.HasValue)
+                {
+                    if (QueryStringAppend)
+                    {
+                        context.HttpContext.Request.QueryString = context.HttpContext.Request.QueryString.Add(query);
```
Tratcher: extract `context.HttpContext.Request` to local variable

src/Microsoft.AspNetCore.Rewrite/Internal/UrlActions/RedirectAction.cs
```diff
@@ -2,49 +2,102 @@
...
-                context.HttpContext.Response.Headers[HeaderNames.Location] = pattern;
+                // If the request url has a query string and the target does not, append the query string
+                // by default.
+                if (QueryStringDelete)
+                {
+                    context.HttpContext.Response.Headers[HeaderNames.Location] = pattern;
+                }
+                else
+                {
+                    context.HttpContext.Response.Headers[HeaderNames.Location] = pattern + context.HttpContext.Request.QueryString;
```
Tratcher: extract `context.HttpContext.Response` to a local variable

Tratcher: in all of these modified files

Tratcher: or consider adding helpers like these: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication/Events/BaseContext.cs#L17-L25

Tratcher:
:shipit:

PR #83, opened 2016-08-31
Removes startsWith in favor of array indexing, More null/empty checks

ZestyBread:
The perf on startsWith was generally slow and I could be using indexing into the string for better perf.
Before
before
After
after

This example was done on RedirectRule, but applies to all rules and actions being applied.

Also there was a case where if the url was an empty string, the rule would not throw on parse, which it should (after investigating IIS).

### DIFF
--- a/src/Microsoft.AspNetCore.Rewrite/Internal/CodeRules/RedirectRule.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/Internal/CodeRules/RedirectRule.cs
@@ -16,6 +16,16 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.CodeRules
         public int StatusCode { get; }
         public RedirectRule(string regex, string replacement, int statusCode)
         {
+            if (string.IsNullOrEmpty(regex))
+            {
+                throw new ArgumentNullException(nameof(regex));
+            }
+
+            if (string.IsNullOrEmpty(replacement))
+            {
+                throw new ArgumentNullException(nameof(replacement));
+            }
+
             InitialMatch = new Regex(regex, RegexOptions.Compiled | RegexOptions.CultureInvariant, _regexTimeout);
             Replacement = replacement;
             StatusCode = statusCode;
@@ -38,9 +48,17 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.CodeRules
             {
                 var newPath = initMatchResults.Result(Replacement);
                 var response = context.HttpContext.Response;
-                response.StatusCode = StatusCode;
 
-                if (newPath.IndexOf("://", StringComparison.Ordinal) == -1 && !newPath.StartsWith("/"))
+                response.StatusCode = StatusCode;
+                context.Result = RuleTermination.ResponseComplete;
+
+                if (string.IsNullOrEmpty(newPath))
+                {
+                    response.Headers[HeaderNames.Location] = "/";
+                    return;
+                }
+
+                if (newPath.IndexOf("://", StringComparison.Ordinal) == -1 && newPath[0] != '/')
                 {
                     newPath = '/' + newPath;
                 }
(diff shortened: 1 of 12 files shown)


ZestyBread:
CC @Tratcher @natemcmaster @davidfowl

src/Microsoft.AspNetCore.Rewrite/Internal/CodeRules/RedirectRule.cs
```diff
@@ -38,9 +48,17 @@ public override void ApplyRule(RewriteContext context)
             {
                 var newPath = initMatchResults.Result(Replacement);
                 var response = context.HttpContext.Response;
+
                 response.StatusCode = StatusCode;
+                context.Result = RuleTermination.ResponseComplete;
+
+                if (string.IsNullOrEmpty(newPath))
+                {
+                    response.Headers[HeaderNames.Location] = string.Empty;
```
Tratcher: This is invalid

Tratcher: You need at least "/:

src/Microsoft.AspNetCore.Rewrite/Internal/CodeRules/RedirectRule.cs
```diff
@@ -38,9 +48,17 @@ public override void ApplyRule(RewriteContext context)
             {
                 var newPath = initMatchResults.Result(Replacement);
                 var response = context.HttpContext.Response;
+
                 response.StatusCode = StatusCode;
+                context.Result = RuleTermination.ResponseComplete;
+
+                if (string.IsNullOrEmpty(newPath))
```
natemcmaster: What leads to `newPath` being null or empty?

src/Microsoft.AspNetCore.Rewrite/Internal/UrlActions/RewriteAction.cs
```diff
@@ -9,7 +9,7 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.UrlActions
 {
     public class RewriteAction : UrlAction
     {
-        private readonly string ForwardSlash = "/";
+        private readonly char ForwardSlash = '/';
```
natemcmaster: You can just use char literals.

src/Microsoft.AspNetCore.Rewrite/Internal/UrlActions/RedirectAction.cs
```diff
@@ -53,7 +53,14 @@ public override void ApplyAction(RewriteContext context, MatchResults ruleMatch,
                 pattern = Uri.EscapeDataString(pattern);
             }
 
-            if (pattern.IndexOf("://", StringComparison.Ordinal) == -1 && !pattern.StartsWith("/"))
+            if (string.IsNullOrEmpty(pattern))
+            {
+                response.Headers[HeaderNames.Location] = string.Empty;
```
natemcmaster: Also not valid. Needs to be a path.

As above, when would this condition trigger? Seems like this might be an error. I want to make sure inferring a redirect to '/' is correct

natemcmaster:
:watch:

src/Microsoft.AspNetCore.Rewrite/Internal/CodeRules/RedirectRule.cs
ZestyBread: It should never be null, but I think the edge case is if a user captures an empty string (which I'm not sure is possible), they can have an empty string.

ZestyBread:
:up: :date:

src/Microsoft.AspNetCore.Rewrite/Internal/CodeRules/RedirectRule.cs
natemcmaster: Since we are speculating on this behavior, could you add a test case to show the valid behavior that hits this path? Otherwise, let's thread this as an error condition instead of redirecting to '/'

ZestyBread:
Added tests that have an empty string there. We do need that check because semantically the empty string and a forward slash should be equal.

natemcmaster:
:shipit:

davidfowl:
👏

PR #89, opened 2016-09-02
Adds path base to redirect location and tests

ZestyBread:
As title says

### DIFF
--- a/src/Microsoft.AspNetCore.Rewrite/Internal/CodeRules/RedirectRule.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/Internal/CodeRules/RedirectRule.cs
@@ -34,6 +34,8 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.CodeRules
         public override void ApplyRule(RewriteContext context)
         {
             var path = context.HttpContext.Request.Path;
+            var pathBase = context.HttpContext.Request.PathBase;
+
             Match initMatchResults;
             if (path == PathString.Empty)
             {
@@ -54,7 +56,7 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.CodeRules
 
                 if (string.IsNullOrEmpty(newPath))
                 {
-                    response.Headers[HeaderNames.Location] = "/";
+                    response.Headers[HeaderNames.Location] = pathBase.HasValue ? pathBase.Value : "/";
                     return;
                 }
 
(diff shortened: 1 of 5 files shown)


ZestyBread:
cc @Tratcher @natemcmaster

src/Microsoft.AspNetCore.Rewrite/Internal/CodeRules/RedirectRule.cs
```diff
@@ -54,7 +56,7 @@ public override void ApplyRule(RewriteContext context)
 
                 if (string.IsNullOrEmpty(newPath))
                 {
-                    response.Headers[HeaderNames.Location] = "/";
+                    response.Headers[HeaderNames.Location] = pathBase + '/';
```
natemcmaster: Do we need to append the '/'?

ZestyBread: Believe so. The pathbase can be null/ empty, so we need to guarantee there is a slash there.

ZestyBread: I could conditionally check on the path base here.

natemcmaster: > pathbase can be null

Seems like `pathBase ?? '/'` would be the right thing in that case

ZestyBread: Its a problem too if it is empty, because we would be setting the location header to string.empty/

natemcmaster: Pretty sure PathString can be null but not empty http://aspnetsource.azurewebsites.net/#Microsoft.AspNetCore.Http.Abstractions/PathString.cs,29

natemcmaster: So...does this work?
```c#
location = pathBase?.HasValue == true ? pathBase.Value : '/';

ZestyBread: I think pathBase cannot be null here, but the value can be.

ZestyBread:
:up: :date:

natemcmaster:
:ship: :it:

Issue #73, opened 2016-08-25
Verify Rewrite middleware for release


muratg:
- Verify different flavors
  - Mod rewrite
  - URL rewrite
  - Code based
-  Missing functionality, and recommendations on adding any of those 
- API surface
- Test coverage
- Etc.

muratg:
cc @ZestyBread @Tratcher

ZestyBread:
@natemcmaster Recommend waiting until Monday after API review/ when changes are merged.

natemcmaster:
Creating a list of issues that we should ensure we follow up on before release:

- [ ] https://github.com/aspnet/BasicMiddleware/issues/94 - if not implemented by release, we should remove ChangeEnvironmentAction and ChangeCookieAction. These are currently stub classes.
- [ ] https://github.com/aspnet/BasicMiddleware/issues/78 - common use case. Ensure our logging reveals enough detail to help users identify what went wrong.
- [ ] https://github.com/aspnet/BasicMiddleware/issues/81 - initial investigation shows that lots of rules can have a big perf hit. We need to provide ways to mitigate this.

natemcmaster:
I've finished some microbenchmarking of the middleware. As we release the first version, it's important we make clear that this middleware should be used judiciously. (cc @danroth27 @DamianEdwards) Processing more than a few dozen rules should not be recommended. High numbers of rules (+100) can drag down the perf of a web app to an almost unusable point.

### Perf investigation results

Here are 5 benchmarks of the latest version of this middleware:
![rewrite_middleware_perf_chart](https://cloud.githubusercontent.com/assets/2696087/18803034/91b6bb5a-81a1-11e6-852d-b76606cb3011.png)

Apache mod_rewrite has a similar perf response. Here are three benchmarks comparing the middleware and mod_rewrite.

![apache_v_middleware](https://cloud.githubusercontent.com/assets/2696087/18803084/fa0382a6-81a1-11e6-9023-f0825b291d96.png)

I tried to obtain a similar comparison with IIS UrlRewrite module. The module performs so well that it did not show any statistically significant degradation of performance as the number of rules increased. IIS UrlRewrite is not a perf bottleneck for ≤ 1000 rules (at least for the scenarios tested in the microbenchmarks).

Tratcher:
These comparisons are all between code rules in the middleware and mod rules in apache, correct? How well does the middleware's implementation of apache's and IIS's rules perform?

natemcmaster:
[...]

PR #97, opened 2016-09-06
Rewrite: add 'Apache' and 'IIS' to internal namespaces and classes

natemcmaster:
Minor code cleanup. No functional changes, just renaming some internal namespaces and classes.

Part of #73 

cc @muratg @Tratcher


Tratcher:
:shipit:

PR #98, opened 2016-09-06
Change param for top-level IIS and Apache extension methods

natemcmaster:
Change IHostingEnvironment parameter to IFileProvider

Also fixup some xmldocs

cc @Tratcher


samples/RewriteSample/Startup.cs
```diff
@@ -11,14 +11,14 @@ namespace RewriteSample
 {
     public class Startup
     {
-        public void Configure(IApplicationBuilder app, IHostingEnvironment hostingEnvironment)
+        public void Configure(IApplicationBuilder app, IHostingEnvironment env)
```
Tratcher: @davidfowl or @DamianEdwards actually asked for hostingEnvironment here

Tratcher: but I guess the templates only use env

src/Microsoft.AspNetCore.Rewrite/ApacheModRewriteOptionsExtensions.cs
```diff
@@ -3,50 +3,45 @@
...
-        /// Imports rules from a mod_rewrite file and adds the rules to current rules. 
+        /// Add rules from an Apache mod_rewrite file
         /// </summary>
-        /// <param name="options">The Rewrite options.</param>
-        /// <param name="hostingEnvironment">The Hosting Environment</param>
+        /// <param name="options">The <see cref="RewriteOptions"/></param>
+        /// <param name="fileProvider">The <see cref="IFileProvider"/> </param>
         /// <param name="filePath">The path to the file containing mod_rewrite rules.</param>
-        public static RewriteOptions AddApacheModRewrite(this RewriteOptions options, IHostingEnvironment hostingEnvironment, string filePath)
+        public static RewriteOptions AddApacheModRewrite(this RewriteOptions options, IFileProvider fileProvider, string filePath)
```
Tratcher: :-1: We didn't like having to pass in IHostingEnvironment and needed to come up with a better idea, but IFileProvider is too generic, it doesn't tell you where to get it.

samples/RewriteSample/Startup.cs
natemcmaster: Primary thinking for this was (1) some may want to pull the rewrite rules from web root instead of content root and (2) I'm starting to lay the foundation for configuring rewrites from `IConfiguration` and/or `Options`.

Tratcher: How about mimicking what we do in Configuration? https://github.com/aspnet/Configuration/blob/dev/src/Microsoft.Extensions.Configuration.FileExtensions/FileConfigurationExtensions.cs#L72

src/Microsoft.AspNetCore.Rewrite/ApacheModRewriteOptionsExtensions.cs
natemcmaster: Open to suggestions. 

> IFileProvider is too generic, it doesn't tell you where to get it.

Regardless, I think we should still provide this API. We can add sugar if we think it's too hard for callers to figure out what to pass in

samples/RewriteSample/Startup.cs
natemcmaster: We would probably need to add a new `RewriteOptionsBuilder` API in that case. I'm not sure semantics of `.SetBasePath` on  `RewriteOptions` are clear

Tratcher: SetConfigFileBasePath?

natemcmaster:
As @Tratcher mentioned, we know we want to remove `IHostingEnvironment` from the `.AddApacheModRewrite` and `.AddIISUrlRewrite` extensions, but an adequate replacement is not obvious. Does either of these options look better? Should we leave as is?

cc @DamianEdwards @muratg @davidfowl  

**option 1** - what's in this PR

```c#
// use IFileProvider instead of IHostingEnvironment
    .AddIISUrlRewrite(hostingEnv.ContentRootFileProvider, "web.config")
    .AddApacheModRewrite(hostingEnv.ContentRootFileProvider, ".htaccess");
```

**option 2** - suggested by @Tratcher 
```c#
// set the IFileProvider on RewriteOptions
    .SetConfigFilePathBase(env.ContentRootPath)
    .AddIISUrlRewrite("web.config")
    .AddApacheModRewrite(".htaccess");
```
Unresolved question: what is the default value of `.SetConfigFileBasePath`?

Tratcher:
(from @davidfowl reading over my shoulder): Number 2 looks cleaner, that's what we do in configuration. Do it.

natemcmaster:
:up: :date:

src/Microsoft.AspNetCore.Rewrite/ApacheModRewriteOptionsExtensions.cs
```diff
@@ -3,50 +3,62 @@
...
-            if (string.IsNullOrEmpty(filePath))
+            if (fileProvider == null)
             {
-                throw new ArgumentException(nameof(filePath));
+                throw new ArgumentNullException(nameof(fileProvider));
             }
 
-            var path = Path.Combine(hostingEnvironment.ContentRootPath, filePath);
-            using (var stream = File.OpenRead(path))
+            var fileInfo = fileProvider.GetFileInfo(filePath);
```
ZestyBread: Should we throw here if the fileInfo doesn't exist?

ZestyBread: Check fileInfo.Exists

natemcmaster: Doesn't really matter. The next line will throw in that situation `fileInfo.CreateReadStream`

src/Microsoft.AspNetCore.Rewrite/RewriteOptions.cs
```diff
@@ -1,24 +1,49 @@
...
 {
     /// <summary>
     /// Options for the <see cref="RewriteMiddleware"/> 
     /// </summary>
-    public class RewriteOptions
+    public class RewriteOptions : IRewriteOptionsInfrastructure
     {
+        public RewriteOptions()
+        {
+            ((IRewriteOptionsInfrastructure)this).DefaultConfigFileProvider = CreateConfigFileProvider();
```
ZestyBread: For my understanding, why do we need to cast here? Thought we can directly access DefaultConfigFileProvider.

natemcmaster: It's a c# thing...explicit interfaces. This obfuscates the `DefaultConfigFileProvider` property from the average user and avoids exposing a `.Internal` API on a public surface.

Tratcher: Overkill, we don't care if users see/set this property.

natemcmaster: I do. It's not actually an 'option' in the sense that it doesn't affect how the middleware operates. It's only there to make options builder APIs work.

Tratcher: EditorBrowsableNever?

src/Microsoft.AspNetCore.Rewrite/RewriteOptionsExtensions.cs
```diff
@@ -12,9 +13,36 @@ namespace Microsoft.AspNetCore.Rewrite
...
+        /// </summary>
+        /// <remarks>
+        /// This should be called before using
+        /// <see cref="ApacheModRewriteOptionsExtensions.AddApacheModRewrite(RewriteOptions, string)"/> or
+        /// <see cref="IISUrlRewriteOptionsExtensions.AddIISUrlRewrite(RewriteOptions, string)"/>.
+        /// </remarks>
+        /// <param name="options">The <see cref="RewriteOptions"/></param>
+        /// <param name="pathBase">The root path to use</param>
+        /// <returns></returns>
+        public static RewriteOptions SetConfigFileBasePath(this RewriteOptions options, string pathBase)
```
Tratcher: would it make more sense to have this take IHostinEnvironment or IFileProvider? Config doesn't do that because they don't exist at that stage, but they will exist when configuring middleware.

natemcmaster: IMO this should hang off an `IRewriteOptionsBuilder` API. This separates builder state from middleware config

src/Microsoft.AspNetCore.Rewrite/RewriteOptions.cs
natemcmaster: Solves the obfuscates  part, but not the '.Internal' part https://github.com/aspnet/BasicMiddleware/pull/98#discussion_r77885612

src/Microsoft.AspNetCore.Rewrite/RewriteOptionsExtensions.cs
ZestyBread: So transfer all of the extension methods to be part of the builder? Seems file.

src/Microsoft.AspNetCore.Rewrite/RewriteOptions.cs
Tratcher: @Eilon this is a style question that may interest you.

natemcmaster: FWIW the explicit interface approach has precendent in [EF options building](https://github.com/aspnet/EntityFramework/blob/03bcb5122e3f577a84498545fcf130ba79a3d987/src/Microsoft.EntityFrameworkCore.InMemory/Extensions/InMemoryDbContextOptionsExtensions.cs#L67) APIs

Eilon: This is quite ugly but I don't care that much 😄

src/Microsoft.AspNetCore.Rewrite/Internal/IRewriteOptionsInfrastructure.cs
```diff
@@ -0,0 +1,15 @@
...
+using Microsoft.Extensions.FileProviders;
+
+namespace Microsoft.AspNetCore.Rewrite.Internal
+{
+    /// <summary>
+    /// Internal API for configuring incognito rewrite options
+    /// </summary>
+    public interface IRewriteOptionsInfrastructure
+    {
+        IFileProvider DefaultConfigFileProvider { get; set; }
```
Tratcher: ConfigFileProvider.

src/Microsoft.AspNetCore.Rewrite/RewriteOptions.cs
```diff
@@ -1,24 +1,49 @@
...
+        }
+
+        protected virtual IFileProvider CreateConfigFileProvider()
+        {
+#if NET451
+            var stringBasePath = AppDomain.CurrentDomain.GetData("APP_CONTEXT_BASE_DIRECTORY") as string
```
Tratcher: This doesn't belong here, this is Hosting's job.

src/Microsoft.AspNetCore.Rewrite/RewriteOptionsExtensions.cs
Tratcher: Looking at this again I think there should only be an IFileProvider version of this method. Then you don't even need the PhysicalFileProvider dependency.

natemcmaster:
Discussed in person with @Tratcher. I'm reverting the PR to represent option 1 (https://github.com/aspnet/BasicMiddleware/pull/98#issuecomment-245120564). The implementation of option2 is introduces state into options that really belongs on a builder API (which we don't have yet and would need to design.)

natemcmaster:
Now that we've had API review, ping for code review.

Tratcher:
:shipit:

PR #99, opened 2016-09-06
Use TaskCache class from Microsoft.Extensions.TaskCache.Sources

pakrym:
Instead of Task.FromResult(0)

### DIFF
--- a/src/Microsoft.AspNetCore.Buffering/project.json
+++ b/src/Microsoft.AspNetCore.Buffering/project.json
@@ -21,7 +21,11 @@
     ]
   },
   "dependencies": {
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Buffering/project.json
```diff
@@ -21,7 +21,8 @@
     ]
   },
   "dependencies": {
-    "Microsoft.AspNetCore.Http.Abstractions": "1.1.0-*"
+    "Microsoft.AspNetCore.Http.Abstractions": "1.1.0-*",
+    "Microsoft.Extensions.TaskCache.Sources": "1.1.0-*"
```
BrennanConroy: Shouldn't this be a sources reference?

PR #101, opened 2016-09-18
IPNetwork Parse/TryParse and common networks

moander:
- Parse and TryParse methods
- Static props for common networks like rfc1918 and link-local


moander:
IPAddress.Parse behaves differently on OS X.

https://github.com/dotnet/corefx/issues/11843

Tratcher:
Please update the samples to show the new APIs

src/Microsoft.AspNetCore.HttpOverrides/IPNetwork.cs
```diff
@@ -1,12 +1,234 @@
...
+
+        /// <summary>
+        /// RFC1918 Private IPv4 range 192.168.0.0/16 (255.255.0.0)
+        /// </summary>
+        public static readonly IPNetwork IPv4Multicast = Parse("224.0.0.0/4");
+
+        /// <summary>
+        /// RFC1918 Private IPv4 Addresses 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16, link-local, and localhost
+        /// </summary>
+        public static readonly IPNetwork[] IPv4PrivateNetworks = new IPNetwork[]
```
nil4: Exposing a `readonly` array allows any consumer to change/overwrite its elements. Is this acceptable here, in `All` and `PrivateNetworks` below?

moander: @nil4: I changed them to ImmutableArray<T> in my latest commit

davidfowl:
We're not going to take a dependency on immutable collections for what seems like a such an obscure use case.

src/Microsoft.AspNetCore.HttpOverrides/project.json
```diff
@@ -24,10 +24,11 @@
   "dependencies": {
     "Microsoft.AspNetCore.Http.Extensions": "1.1.0-*",
     "Microsoft.Extensions.Logging.Abstractions": "1.1.0-*",
-    "Microsoft.Extensions.Options": "1.1.0-*"
+    "Microsoft.Extensions.Options": "1.1.0-*",
+    "System.Collections.Immutable": "1.2.0"
```
davidfowl: Yea, nope.

moander: Please elaborate :)

davidfowl: Make it a readonly list. We're not going to take a dependency to expose a not so common exchange type.

moander: Are there some reference-policy-guidelines you are following or did you decide that one on your own?

davidfowl: Decided it on my own based on the fact that we don't expose immutable* anywhere else in all of asp.net. Also decided it based on the fact that interfaces should be preferred over concrete types like this. It's the same reason we don't expose List in public API. There's probably a guideline somewhere though, but you can take my word for it 😆

src/Microsoft.AspNetCore.HttpOverrides/IPNetwork.cs
```diff
@@ -1,12 +1,245 @@
...
+        /// <summary>
+        /// IPv4 Multicast range 224.0.0.0/4
+        /// </summary>
+        public static readonly IPNetwork IPv4Multicast = Parse("224.0.0.0/4");
+
+        /// <summary>
+        /// RFC1918 Private IPv4 Addresses 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16
+        /// </summary>
+        /// <seealso cref="IPv4PrivateNetworks" />
+        public static readonly ImmutableArray<IPNetwork> RFC1918 = ImmutableArray.Create(IPv4Private10, IPv4Private172, IPv4Private192);
```
Tratcher: The name is precise, but not descriptive. Give it a descriptive name and put the RFC in the doc comments.

src/Microsoft.AspNetCore.HttpOverrides/IPNetwork.cs
```diff
@@ -1,12 +1,245 @@
...
+        /// </summary>
+        /// <seealso cref="PrivateNetworks" />
+        /// <seealso cref="IPv4PrivateNetworks" />
+        /// <seealso cref="IPv6PrivateNetworks" />
+        public static readonly IPNetwork IPv6Private = Parse("fc00::/7");
+
+        /// <summary>
+        /// IPv6 Teredo range 2001::/32
+        /// </summary>
+        public static readonly IPNetwork Teredo = Parse("2001::/32");
```
Tratcher: Let's restrict the constants to the ones that will be commonly used in the forwarded headers scenario. I would not expect Teredo, Multicast, etc. to be used in these scenarios.

src/Microsoft.AspNetCore.HttpOverrides/IPNetwork.cs
```diff
@@ -1,12 +1,245 @@
...
+
+        /// <summary>
+        /// All IPv4 addresses 0.0.0.0/0
+        /// </summary>
+        public static readonly IPNetwork AllIPv4 = Parse("0.0.0.0/0");
+
+        /// <summary>
+        /// IPv4 localhost range 127.0.0.0/8 (255.0.0.0)
+        /// </summary>
+        public static readonly IPNetwork IPv4Loopback = Parse("127.0.0.1/8");
```
Tratcher: Add this to the sample to show usage

Tratcher: And update the default value of ForwardedHeadersOptions.KnownNetworks to use this static

src/Microsoft.AspNetCore.HttpOverrides/IPNetwork.cs
```diff
@@ -1,12 +1,245 @@
...
+            IPv4LinkLocal,
+            IPv4Private10,
+            IPv4Private172,
+            IPv4Private192,
+            IPv6Loopback,
+            IPv6LinkLocal,
+            IPv6Private);
+
+        // Helper method for Parse/TryParse
+        private static bool TryParsePrefixLength(IPAddress prefix, string[] parts, out byte length)
```
Tratcher: style: methods go after the constructor, even static methods.

src/Microsoft.AspNetCore.HttpOverrides/IPNetwork.cs
```diff
@@ -63,5 +296,10 @@ private byte[] CreateMask()
 
             return mask;
         }
+
+        public override string ToString()
+        {
+            return Prefix.ToString() + "/" + PrefixLength;
```
Tratcher: PrefixLength.ToString(CultureInvariant)

src/Microsoft.AspNetCore.HttpOverrides/IPNetwork.cs
```diff
@@ -1,12 +1,245 @@
...
+            byte length;
+            if (!TryParsePrefixLength(prefix, parts, out length))
+            {
+                throw new FormatException("Invalid prefix length");
+            }
+
+            return new IPNetwork(prefix, length);
+        }
+
+        public IPNetwork(IPAddress prefix)
```
Tratcher: This constructor doesn't make sense for use with ForwardedHeadersOptions (and you never call it). Single addresses go on KnownProxies.

src/Microsoft.AspNetCore.HttpOverrides/IPNetwork.cs
```diff
@@ -1,12 +1,245 @@
...
+            IPv6Loopback,
+            IPv6LinkLocal,
+            IPv6Private);
+
+        // Helper method for Parse/TryParse
+        private static bool TryParsePrefixLength(IPAddress prefix, string[] parts, out byte length)
+        {
+            if (parts.Length > 1)
+            {
+                return byte.TryParse(parts[1], NumberStyles.Number, null, out length)
```
Tratcher: InvariantCulture

src/Microsoft.AspNetCore.HttpOverrides/project.json
davidfowl: So lets change it to IReadOnlyList and remove the dependency on Immutable collections.

muratg:
@moander Could you sign the CLA if you agree with the changes proposed?

muratg:
Closed/reopend to trigger the CLA bot

src/Microsoft.AspNetCore.HttpOverrides/IPNetwork.cs
Tratcher: Remove all the constants except maybe IPv4Loopback and IPv6Loopback. Most of these are impractical in the context of forwarder filters.

test/Microsoft.AspNetCore.HttpOverrides.Tests/IPNetworkTest.cs
```diff
@@ -28,5 +29,73 @@ public void Contains_Negative(string prefixText, int length, string addressText)
...
+            network = IPNetwork.Parse(networkString);
+            Assert.Equal(expectedPrefix, network.Prefix.ToString());
+            Assert.Equal(expectedLength, network.PrefixLength);
+        }
+
+        [Theory]
+        [InlineData("foo")]
+        // TODO: This test fails on OS X
+        // See https://github.com/dotnet/corefx/issues/11843
+        [InlineData("4294967296")]
```
Tratcher: Comment out this test case then

moander: Are you sure you don't want this test should pass before merging the PR? The test fails because the parser  gives  different on OS X compared to Linux and Windows.

Tratcher: Yes, it should be fine. When they fix the bug we won't need to make any further changes, and there's no telling how long it will take them.

muratg:
@moander, thanks for sending this PR. Closing this one as this is quite stale right now. Please file a bug if you'd like to bring this back to discussion.

PR #102, opened 2016-09-18
Forward custom header names

moander:
Made it possible to use custom header names for X-Forwarded-For|Host|Proto

### DIFF
--- a/src/Microsoft.AspNetCore.HttpOverrides/ForwardedHeadersDefaults.cs
+++ b/src/Microsoft.AspNetCore.HttpOverrides/ForwardedHeadersDefaults.cs
@@ -12,31 +12,31 @@ namespace Microsoft.AspNetCore.HttpOverrides
         /// <summary>
         /// X-Forwarded-For
         /// </summary>
-        public const string XForwardedForHeaderName = "X-Forwarded-For";
+        public static string XForwardedForHeaderName { get; } = "X-Forwarded-For";
 
         /// <summary>
         /// X-Forwarded-Host
         /// </summary>
-        public const string XForwardedHostHeaderName = "X-Forwarded-Host";
+        public static string XForwardedHostHeaderName { get; } = "X-Forwarded-Host";
 
         /// <summary>
         /// X-Forwarded-Proto
         /// </summary>
-        public const string XForwardedProtoHeaderName = "X-Forwarded-Proto";
+        public static string XForwardedProtoHeaderName { get; } = "X-Forwarded-Proto";
 
         /// <summary>
         /// X-Original-For
         /// </summary>
-        public const string XOriginalForHeaderName = "X-Original-For";
+        public static string XOriginalForHeaderName { get; } = "X-Original-For";
 
         /// <summary>
         /// X-Original-Host
         /// </summary>
-        public const string XOriginalHostHeaderName = "X-Original-Host";
+        public static string XOriginalHostHeaderName { get; } = "X-Original-Host";
 
         /// <summary>
         /// X-Original-Proto
         /// </summary>
-        public const string XOriginalProtoHeaderName = "X-Original-Proto";
+        public static string XOriginalProtoHeaderName { get; } = "X-Original-Proto";
     }
 }
--- a/src/Microsoft.AspNetCore.HttpOverrides/ForwardedHeadersDefaults.cs
+++ b/src/Microsoft.AspNetCore.HttpOverrides/ForwardedHeadersDefaults.cs
@@ -1,0 +1,42 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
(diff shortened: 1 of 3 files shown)


Tratcher:
What other names did you plan to use? Are you sure the semantics of those other headers are the same?

moander:
The developer should be allowed to decide what other names to use, and also be responsible for making sure the semantics of those are the same. I agree that my examples are irresponsible as they have not been tested properly.

davidfowl:
> The developer should be allowed to decide what other names to use, and also be responsible for making sure the semantics of those are the same. I agree that my examples are irresponsible as they have not been tested properly.

Why? What real scenario inspired this change in the first place?

moander:
Google App Engine

davidfowl:
What names do they use?

moander:
X-Google-Real-IP

Try to search for ` x-forwarded-for` on Google if you want more examples. There are plenty :)

Here is one example found in the HAProxy docs:

The keyword "header" may be used to supply a different header name to replace the default "X-Forwarded-For". This can be useful where you might already have a "X-Forwarded-For" header from a different application (eg: stunnel), and you need preserve it. Also if your backend server doesn't use the "X-Forwarded-For" header and requires different one (eg: Zeus Web Servers require "X-Cluster-Client-IP").

davidfowl:
I don't have any pushback against the feature. It just needs to exist for good reasons.

src/Microsoft.AspNetCore.HttpOverrides/ForwardedHeadersMiddleware.cs
```diff
@@ -17,9 +17,18 @@ namespace Microsoft.AspNetCore.HttpOverrides
...
-        private const string XForwardedProtoHeaderName = "X-Forwarded-Proto";
+        private string XForwardedForHeaderName
+        {
+            get { return _options.ForwardedForHeaderName; }
+        }
+        private string XForwardedHostHeaderName
+        {
+            get { return _options.ForwardedHostHeaderName; }
+        }
+        private string XForwardedProtoHeaderName
```
Tratcher: style: either remove these and access options directly, or at least move them down after the constructor

src/Microsoft.AspNetCore.HttpOverrides/ForwardedHeadersOptions.cs
```diff
@@ -10,6 +10,38 @@ namespace Microsoft.AspNetCore.Builder
...
+
+        /// <summary>
+        /// X-Forwarded-Proto
+        /// </summary>
+        public const string DefaultForwardedProtoHeaderName = "X-Forwarded-Proto";
+
+        /// <summary>
+        /// Use this header name instead of X-Forwarded-For
+        /// </summary>
+        public string ForwardedForHeaderName { get; set; } = DefaultForwardedForHeaderName;
```
Tratcher: In the ForwardedHeadersMiddleware constructor validate that these are not null or empty

src/Microsoft.AspNetCore.HttpOverrides/ForwardedHeadersOptions.cs
```diff
@@ -10,6 +10,38 @@ namespace Microsoft.AspNetCore.Builder
     public class ForwardedHeadersOptions
     {
         /// <summary>
+        /// X-Forwarded-For
+        /// </summary>
+        public const string DefaultForwardedForHeaderName = "X-Forwarded-For";
```
Tratcher: We have precedence for separating these to their own class.
https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationDefaults.cs
How about ForwardedHeaderNames.XForwardedFor, etc..

src/Microsoft.AspNetCore.HttpOverrides/ForwardedHeadersMiddleware.cs
troydai: style: also you can use a more concise form of get-only property
```
private string XForwardedForHeaderName => _options.ForwardedForHeaderName;
```

src/Microsoft.AspNetCore.HttpOverrides/ForwardedHeadersDefaults.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// <summary>
+    /// Default values related to <see cref="ForwardedHeadersMiddleware"/> middleware
+    /// </summary>
+    /// <seealso cref="Microsoft.AspNetCore.Builder.ForwardedHeadersOptions"/>
+    public static class ForwardedHeadersDefaults
+    {
+        /// <summary>
+        /// X-Forwarded-For
+        /// </summary>
+        public const string XForwardedForHeaderName = "X-Forwarded-For";
```
Tratcher: Make these `public static readonly`. We avoid `public const` APIs.

Tratcher:
Thanks for being responsive.

moander:
@Tratcher Thanks for reviewing :)

Issue #94, opened 2016-09-02
Add cookie and environment variable actions


jkotalik:
mod_rewrite specific thing. These flags modify the cookie and environment variable respectively.
See:
https://github.com/aspnet/BasicMiddleware/blob/dev/src/Microsoft.AspNetCore.Rewrite/Internal/UrlActions/ChangeCookieAction.cs

muratg:
Environment variable actions is not possible with our current implementation -- they're set per process.

Cookie ones could be done.

muratg:
Assigning to @natemcmaster to remove the Environment one. We'll either implement or remove Cookie one before 1.1.

PR #107, opened 2016-09-23
Remove ChangeEnvironmentAction class

natemcmaster:
See https://github.com/aspnet/BasicMiddleware/issues/94#issuecomment-249297134

cc @mikaelm12


src/Microsoft.AspNetCore.Rewrite/Internal/ApacheModRewrite/RuleBuilder.cs
```diff
@@ -176,12 +176,6 @@ public void AddRule(string rule)
                 _actions.Add(new ChangeCookieAction(flag));
             }
 
-            if (flags.GetValue(FlagType.Env, out flag))
-            {
-                // parse env
-                _actions.Add(new ChangeEnvironmentAction(flag));
```
Tratcher: NotSupportedException?

natemcmaster: It doesn't seem we throw this when other unsupported flags are used. We just silently ignore...should we open a separate issue for that?

Tratcher: In most places we did throw either when parsing the flag or where it would have been implemented. Anything we can do to throw an error at startup telling them their rules contain unsupported features would be good.

natemcmaster: Yeah, I can totally agree with that approach. Will add a throw.

mikaelm12:
Looks fine. So this isn't going in 1.1?

jkotalik:
This really was only a notion is mod_rewrite, and frankly this doesn't seem like a middleware that should be responsible for changing the environment variable.

natemcmaster:
:up: :date:

> this isn't going in 1.1?

It's not going in at all. The feature isn't possible.

src/Microsoft.AspNetCore.Rewrite/Internal/ApacheModRewrite/RuleBuilder.cs
```diff
@@ -176,10 +176,10 @@ public void AddRule(string rule)
                 _actions.Add(new ChangeCookieAction(flag));
             }
 
-            if (flags.GetValue(FlagType.Env, out flag))
+            string envFlag;
+            if (flags.GetValue(FlagType.Env, out envFlag))
```
Tratcher: revert diff

Issue #108, opened 2016-09-26
Clean up response compression


Tratcher:
Open items:
- More convenient Options
 - ~~Compression providers~~
 - ~~Compression negotiator~~
- accept-encoding q-value negotiation (perf)
- ~~should mime types be case sensitive? (No)~~
- ~~Make sure no negotiation happens / no bytes written if there's no body.~~
- ~~Flush~~
- Exception
- ~~Hook ISendFileFeature?~~
- API Review

Testing:
- ~~How much buffering does the gzip stream do?~~ 8kb
- ~~Integration test with WebListener, Kestrel,~~
- ~~IIS/Express~~
- ~~nginx~~ https://github.com/aspnet/BasicMiddleware/issues/123
- ~~Perf~~ https://github.com/dotnet/corefx/issues/12549

Release:
- [x] Bump version from 0.1.0 to 1.0.0 when ready

muratg:
Putting in 1.1.0 for now. We can move it to 1.2.0 if we can't finish this off in time.

Tratcher:
GZipStream allocations: https://github.com/dotnet/corefx/issues/12549

muratg:
@Tratcher Per PM decision, version numbers for new packages will be 1.0.0-preview1 for the initial release. (Not 1.1.0-preview1)

Tratcher:
Review feedback:

Remove the `AddResponseCompression(params string[])` overload, investigate defaults and wildcards instead. Text/*, Application/javascript

Try binding Options from Config

Should the services be registered as singleton or transient? Eilon said Singleton.

Split ResponseCompressionExtensions by services vs appbuilder.

Rename EnableHttps to EnableForHttps

Tratcher:
Reading from config works pretty well:
```
{
  "section": {
    "EnableForHttps": true,
    "MimeTypes": ["text/plain", "text/html"]
  }
}
```
```
            services.AddResponseCompression(options =>
            {
                Configuration.GetSection("section").Bind(options);
            });
```

Tratcher:
Possible default mime types (things used by the template, static files):
text/css
text/html
application/javascript

WebApi:
application/xml
text/xml
application/json
text/json
application/json-patch+json

Other:
text/plain

Yves57:
Maybe adding "image/svg+xml" to the default types.

Tratcher:
[...]

PR #113, opened 2016-09-29
Change ResponseCompression to be DI centric

Tratcher:
#108 DI all the things
@JunTaoLuo @davidfowl @Yves57 

The only behavioral change is that I removed the multiple ContentEncodings restriction.


samples/ResponseCompressionSample/Startup.cs
```diff
@@ -5,21 +5,26 @@
...
 using Microsoft.AspNetCore.ResponseCompression;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace ResponseCompressionSample
 {
     public class Startup
     {
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddTransient<ICompressionProvider, GzipCompressionProvider>();
```
Yves57: Why "Transient"?

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionOptions.cs
```diff
@@ -13,19 +11,13 @@ namespace Microsoft.AspNetCore.ResponseCompression
     public class ResponseCompressionOptions
     {
         /// <summary>
-        /// Called when an HTTP request accepts a compatible compression algorithm, and returns True
-        /// if the response should be compressed.
+        /// Response Content-Type MIME types to compress.
         /// </summary>
-        public Func<HttpContext, bool> ShouldCompressResponse { get; set; }
+        public IEnumerable<string> MimeTypes { get; set; }
```
Yves57: Just for curiosity, why back to "MimeTypes"?

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionProvider.cs
```diff
@@ -0,0 +1,102 @@
...
+            if (_providers.Length == 0)
+            {
+                _providers = new [] { new GzipCompressionProvider() };
+            }
+
+            if (options.Value.MimeTypes == null || !options.Value.MimeTypes.Any())
+            {
+                throw new InvalidOperationException("No mime types specified");
+            }
+            _mimeTypes = new HashSet<string>(options.Value.MimeTypes, StringComparer.OrdinalIgnoreCase);
```
Yves57: So it will create a new `HashSet<>` for each request? No perf problems?

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionProvider.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// If no compression providers are specified then GZip is used by default.
+        /// </summary>
+        /// <param name="providers">Compression providers to use, if any.</param>
+        /// <param name="options"></param>
+        public ResponseCompressionProvider(IEnumerable<ICompressionProvider> providers, IOptions<ResponseCompressionOptions> options)
+        {
+            _providers = providers.ToArray();
+            if (_providers.Length == 0)
+            {
+                _providers = new [] { new GzipCompressionProvider() };
```
Yves57: Is it not better to call `TryAddTransient<ICompressionProvider, GzipCompressionProvider>()` (or `TryAddSingleton`) in `AddResponseCompression`?

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionProvider.cs
```diff
@@ -0,0 +1,102 @@
...
+                    {
+                        return _providers[0];
+                    }
+                    if (string.Equals("identity", encoding.Value, StringComparison.OrdinalIgnoreCase))
+                    {
+                        return null;
+                    }
+
+                    // There will rarely be more than three providers, and there's only one by default
+                    foreach (var provider in _providers)
```
Yves57: At least IE Edge or Chrome seem to fill AcceptEncoding with explicity values.
So I think it's probably faster to check `_providers` before "*" and "identity".

samples/ResponseCompressionSample/Startup.cs
Tratcher: Transient because the class is stateless, DI doesn't need to hold onto it.

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionOptions.cs
Tratcher: Because ShouldCompressResponse moved to IResponseCompressionProvider and this was the only sane way I could find to flow mime types into there.

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionProvider.cs
Tratcher: No, ResponseCompressionProvider is created once for the middleware.

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionProvider.cs
Tratcher: That may lead to duplicates depending on the order the app registers services.

Tratcher:
Updated

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionExtensions.cs
```diff
@@ -13,22 +14,53 @@ namespace Microsoft.AspNetCore.Builder
...
+            return collection.AddResponseCompression(options =>
+            {
+                options.MimeTypes = mimeTypes;
+            });
+        }
+
+        /// <summary>
+        /// Add response compression services and configure the related options.
+        /// </summary>
+        /// <param name="collection"></param>
```
JunTaoLuo: nit: we usually call it `services` instead of `collection`? Also, elaborate on the params and return types of the xml docs?

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionMiddleware.cs
```diff
@@ -20,45 +14,20 @@ public class ResponseCompressionMiddleware
...
         private readonly bool _enableHttps;
 
         /// <summary>
         /// Initialize the Response Compression middleware.
         /// </summary>
         /// <param name="next"></param>
+        /// <param name="provider"></param>
         /// <param name="options"></param>
-        public ResponseCompressionMiddleware(RequestDelegate next, IOptions<ResponseCompressionOptions> options)
+        public ResponseCompressionMiddleware(RequestDelegate next, IResponseCompressionProvider provider, IOptions<ResponseCompressionOptions> options)
```
JunTaoLuo: Add null checks?

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+    {
+        private readonly ICompressionProvider[] _providers;
+        private readonly HashSet<string> _mimeTypes;
+
+        /// <summary>
+        /// If no compression providers are specified then GZip is used by default.
+        /// </summary>
+        /// <param name="providers">Compression providers to use, if any.</param>
+        /// <param name="options"></param>
+        public ResponseCompressionProvider(IEnumerable<ICompressionProvider> providers, IOptions<ResponseCompressionOptions> options)
```
JunTaoLuo: null checks?

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+                var sorted = unsorted
+                    .Where(s => s.Quality.GetValueOrDefault(1) > 0)
+                    .OrderByDescending(s => s.Quality.GetValueOrDefault(1));
+
+                foreach (var encoding in sorted)
+                {
+                    // There will rarely be more than three providers, and there's only one by default
+                    foreach (var provider in _providers)
+                    {
+                        if (string.Equals(provider.EncodingName, encoding.Value, StringComparison.OrdinalIgnoreCase))
```
JunTaoLuo: Just curious how to resolve the same provider for multiple encodings? For example, I want to use the gzip stream for both text/plain and text/html? Do I need to add two compression providers that only differ by the encoding name?

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+        public virtual bool ShouldCompressResponse(HttpContext context)
+        {
+            var mimeType = context.Response.ContentType;
+
+            if (string.IsNullOrEmpty(mimeType))
+            {
+                return false;
+            }
+
+            var separator = mimeType.IndexOf(';');
```
JunTaoLuo: use `StringSegment` to compare for perf?

Tratcher: I'll do a perf pass later

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionProvider.cs
Tratcher: The encoding name is gzip, it's unrelated to the mime type (text/plain)

JunTaoLuo: ah yes, I got mixed up.

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionProvider.cs
Yves57: Yes, sorry!

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+            {
+                // TODO PERF: clients don't usually include quality values so this sort will not have any effect. Fast-path?
+                var sorted = unsorted
+                    .Where(s => s.Quality.GetValueOrDefault(1) > 0)
+                    .OrderByDescending(s => s.Quality.GetValueOrDefault(1));
+
+                foreach (var encoding in sorted)
+                {
+                    // There will rarely be more than three providers, and there's only one by default
+                    foreach (var provider in _providers)
```
Yves57: `for (var i = ....)` and not `foreach` (`_provider`is an array).

Tratcher: For perf, right? I've actually had foreach show up as more efficient in recent profiles elsewhere. I'll profile this code more closely soon.

Yves57: Yes, even if I suppose it is not so important.
A surprise for me because I was thinking that creating a enumerator was slower than manipulating a index.
For information,  @davidfowl approved a similar change in another PR (see it last comment in the [PR](https://github.com/aspnet/HttpAbstractions/pull/709)).

Tratcher: Yeah, never assume perf, always measure. I'll be doing that soon.

samples/ResponseCompressionSample/CustomCompressionProvider.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.IO;
+using Microsoft.AspNetCore.ResponseCompression;
+
+namespace ResponseCompressionSample
+{
+    public class CustomCompressionProvider : ICompressionProvider
+    {
+        public string EncodingName => "custom";
+
+        public Stream CreateStream(Stream outputStream)
```
davidfowl: Need more @pranavkm  points.

```C#
Stream CreateStream(Stream output) => output;
```

samples/ResponseCompressionSample/Startup.cs
davidfowl: Is it resolved once per request?

Tratcher: No, one when constructing the middleware/provider.

davidfowl: Transient is fine (and probably better 😄 )

PR #116, opened 2016-10-04
Removing substring calls

mikaelm12:
In response to issue [#80](https://github.com/aspnet/BasicMiddleware/issues/80)
@natemcmaster @Tratcher @muratg

### DIFF
--- a/samples/RewriteSample/Startup.cs
+++ b/samples/RewriteSample/Startup.cs
@@ -15,7 +15,7 @@ namespace RewriteSample
         {
             var options = new RewriteOptions()
                 .AddRedirect("(.*)/$", "$1")
-                .AddRewrite(@"app/(\d+)", "app?id=$1", skipRemainingRules: false)
+                .AddRewrite(@"^/app/(\d+)", "app?id=$1", skipRemainingRules: false)
                 .AddRedirectToHttps(302, 5001)
                 .AddIISUrlRewrite(env.ContentRootFileProvider, "UrlRewrite.xml")
                 .AddApacheModRewrite(env.ContentRootFileProvider, "Rewrite.txt");
--- a/samples/RewriteSample/Startup.cs
+++ b/samples/RewriteSample/Startup.cs
@@ -14,7 +14,7 @@ namespace RewriteSample
         public void Configure(IApplicationBuilder app, IHostingEnvironment env)
         {
             var options = new RewriteOptions()
-                .AddRedirect("(.*)/$", "$1")
+                .AddRedirect("^/(.*)/$", "$1")
                 .AddRewrite(@"^/app/(\d+)", "app?id=$1", skipRemainingRules: false)
                 .AddRedirectToHttps(302, 5001)
                 .AddIISUrlRewrite(env.ContentRootFileProvider, "UrlRewrite.xml")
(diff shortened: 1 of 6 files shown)


Tratcher:
@jkotalik ? There was a reason for this. Something about how none of the providers wanted you give the leading slash.

samples/RewriteSample/Startup.cs
```diff
@@ -14,8 +14,8 @@ public class Startup
         public void Configure(IApplicationBuilder app, IHostingEnvironment env)
         {
             var options = new RewriteOptions()
-                .AddRedirect("(.*)/$", "$1")
-                .AddRewrite(@"app/(\d+)", "app?id=$1", skipRemainingRules: false)
+                .AddRedirect("^/(.*)/$", "$1")
+                .AddRewrite(@"^/app/(\d+)", "app?id=$1", skipRemainingRules: false)
```
natemcmaster: Is behavior the same without the leading slash?

test/Microsoft.AspNetCore.Rewrite.Tests/MiddlewareTests.cs
```diff
@@ -16,7 +16,7 @@ public class MiddlewareTests
         [Fact]
         public async Task CheckRewritePath()
         {
-            var options = new RewriteOptions().AddRewrite("(.*)", "http://example.com/$1", skipRemainingRules: false);
+            var options = new RewriteOptions().AddRewrite("(.*)", "http://example.com$1", skipRemainingRules: false);
```
natemcmaster: Does this is affect parsed mod_rewrite/URL Rewrite rules? e.g. `RewriteRule (.*) http://example.com/$1`

src/Microsoft.AspNetCore.Rewrite/Internal/RedirectRule.cs
```diff
@@ -35,16 +35,7 @@ public virtual void ApplyRule(RewriteContext context)
...
-            Match initMatchResults;
-            if (path == PathString.Empty)
-            {
-                initMatchResults = InitialMatch.Match(path.ToString());
-            }
-            else
-            {
-                initMatchResults = InitialMatch.Match(path.ToString().Substring(1));
-            }
+            var  initMatchResults = InitialMatch.Match(path.ToString());
```
natemcmaster: nit: extra space between var and initMatchResults

samples/RewriteSample/Startup.cs
natemcmaster: Won't this break the expected behavior of rules parsed from mod_rewrite and URLRewrite?

mikaelm12: Just from some basic testing with variations of the sample things seem to work fine

jkotalik: Be super careful with the leading slashes. Double check both IIS and mod_rewrite for expectations on them, because IIRC they both explicitly disallow leading slashes.

jkotalik:
We need to have consistency between all three implementations. This, I believe, breaks it.
For example:
https://www.iis.net/learn/extensions/url-rewrite-module/creating-rewrite-rules-for-the-url-rewrite-module

<rewrite>
  <rules>
    <rule name="Rewrite to article.aspx">
      <match url="^article/([0-9]+)/([_0-9a-z-]+)" />
      <action type="Rewrite" url="article.aspx?id={R:1}&amp;title={R:2}" />
    </rule>
  </rules>
</rewrite>

IIS rules explicitly match on no leading slash, as well as mod_rewrite. Changing it to require leading slashes for normal rules removes consistency.

Tratcher:
Yeah, let's close this.

jkotalik:
Some of the changes here for style are good though

mikaelm12:
Alright. I'll scrap this then

PR #122, opened 2016-10-07
Added more logging for rewrite/redirect scenarios

mikaelm12:
Logging for 
- Redirect to HTTPS
- The current url when continuing to the next url
- The final url when no longer applying rules

@natemcmaster @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Rewrite/RewriteMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/RewriteMiddleware.cs
@@ -72,21 +72,24 @@ namespace Microsoft.AspNetCore.Rewrite
                 Result = RuleResult.ContinueRules
             };
 
+            context.Items.Add("originalUrl", $"{context.Request.Path + context.Request.QueryString}");
             foreach (var rule in _options.Rules)
             {
                 rule.ApplyRule(rewriteContext);
+                var currentUrl = $"{context.Request.Path + context.Request.QueryString}";
                 switch (rewriteContext.Result)
                 {
                     case RuleResult.ContinueRules:
-                        _logger.RewriteMiddlewareRequestContinueResults();
+                        _logger.RewriteMiddlewareRequestContinueResults(currentUrl);
                         break;
                     case RuleResult.EndResponse:
+                        //_logger.RewriteMiddlewareRequestStopRules($"{context.Request.Path + context.Request.QueryString}");
                         _logger.RewriteMiddlewareRequestResponseComplete(
                             context.Response.Headers[HeaderNames.Location],
                             context.Response.StatusCode);
                         return TaskCache.CompletedTask;
                     case RuleResult.SkipRemainingRules:
-                        _logger.RewriteMiddlewareRequestStopRules();
+                        _logger.RewriteMiddlewareRequestStopRules(currentUrl);
                         return _next(context);
                     default:
                         throw new ArgumentOutOfRangeException($"Invalid rule termination {rewriteContext.Result}");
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNetCore.Rewrite/RewriteMiddleware.cs
```diff
@@ -75,18 +75,19 @@ public Task Invoke(HttpContext context)
             foreach (var rule in _options.Rules)
             {
                 rule.ApplyRule(rewriteContext);
+                var currentUrl = $"{context.Request.Path + context.Request.QueryString}";
```
Tratcher: perf: you only want to do this if logging is enabled

src/Microsoft.AspNetCore.Rewrite/RewriteMiddleware.cs
```diff
@@ -75,18 +75,23 @@ public Task Invoke(HttpContext context)
             foreach (var rule in _options.Rules)
             {
                 rule.ApplyRule(rewriteContext);
+                var currentUrl = string.Empty;
+                if (rewriteContext.Logger.IsEnabled(LogLevel.Debug))
+                {
+                    currentUrl = $"{context.Request.Path + context.Request.QueryString}";
+                }
```
jkotalik: nit: newline probable nice here

src/Microsoft.AspNetCore.Rewrite/RewriteMiddleware.cs
```diff
@@ -75,18 +75,23 @@ public Task Invoke(HttpContext context)
             foreach (var rule in _options.Rules)
             {
                 rule.ApplyRule(rewriteContext);
+                var currentUrl = string.Empty;
+                if (rewriteContext.Logger.IsEnabled(LogLevel.Debug))
```
jkotalik: Need to check if Logger is null here

src/Microsoft.AspNetCore.Rewrite/RewriteMiddleware.cs
```diff
@@ -75,18 +75,24 @@ public Task Invoke(HttpContext context)
             foreach (var rule in _options.Rules)
             {
                 rule.ApplyRule(rewriteContext);
+                var currentUrl = string.Empty;
+
+                if (rewriteContext.Logger?.IsEnabled(LogLevel.Debug) == true)
+                {
+                    currentUrl = $"{context.Request.Path + context.Request.QueryString}";
```
natemcmaster: String interpolation is overkill here.

natemcmaster:
As long as you're at it, can we add an information level log message that summarizes the rewrite in one line?

e.g. "Request was rewritten to {newUrl}",

src/Microsoft.AspNetCore.Rewrite/RewriteMiddleware.cs
```diff
@@ -77,10 +77,11 @@ public Task Invoke(HttpContext context)
                 rule.ApplyRule(rewriteContext);
                 var currentUrl = string.Empty;
 
-                if (rewriteContext.Logger?.IsEnabled(LogLevel.Debug) == true)
+                if (rewriteContext.Logger?.IsEnabled(LogLevel.Debug | LogLevel.Information) == true)
```
natemcmaster: Use `.IsEnabled(LogLevel.Debug) || logger.IsEnabled(LogLevel.Information)` instead. LogLevel isn't designed to be used as bit flags.

src/Microsoft.AspNetCore.Rewrite/RewriteMiddleware.cs
```diff
@@ -97,6 +98,7 @@ public Task Invoke(HttpContext context)
                     default:
                         throw new ArgumentOutOfRangeException($"Invalid rule termination {rewriteContext.Result}");
                 }
+                _logger.RewriteSummary(currentUrl);
```
natemcmaster: It's possible this logger statement won't be produced if SkipRemainingRules is true. It's also possible that this summary is inaccurate. This should instead go inside the `IRule` implementations that actually rewrite the URL.

PR #125, opened 2016-10-12
compression code cleanup

Tratcher:
https://github.com/aspnet/BasicMiddleware/issues/108#issuecomment-253042371 @Eilon @DamianEdwards @JunTaoLuo 

Fixing several minor items from the API review.
- Use Singleton services rather than Transient
- EnableHttps -> EnableForHttps
- Split ResponseCompressionExtensions

The larger items will be addressed separately

### DIFF
--- a/src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionProvider.cs
+++ b/src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionProvider.cs
@@ -41,7 +41,7 @@ namespace Microsoft.AspNetCore.ResponseCompression
 
             if (options.Value.MimeTypes == null || !options.Value.MimeTypes.Any())
             {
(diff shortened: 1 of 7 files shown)


JunTaoLuo:
:shipit:

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionProvider.cs
```diff
@@ -41,7 +41,7 @@ public ResponseCompressionProvider(IEnumerable<ICompressionProvider> providers,
 
             if (options.Value.MimeTypes == null || !options.Value.MimeTypes.Any())
             {
-                throw new InvalidOperationException("No mime types specified");
+                throw new InvalidOperationException("No MIME types specified");
```
Eilon: End sentences with a period

PR #126, opened 2016-10-12
Move compression providers back to options

Tratcher:
https://github.com/aspnet/BasicMiddleware/issues/108#issuecomment-253042371 @DamianEdwards @Eilon @rynowak 
Moved ICompressionProvider registration from DI to options. Emulates MVC's FilterCollection.

### DIFF
--- a/src/Microsoft.AspNetCore.ResponseCompression/CompressionProviderCollection.cs
+++ b/src/Microsoft.AspNetCore.ResponseCompression/CompressionProviderCollection.cs
@@ -1,0 +1,51 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.ObjectModel;
+#if NETSTANDARD1_3
+using System.Reflection;
+#endif
+
+namespace Microsoft.AspNetCore.ResponseCompression
+{
+    /// <summary>
+    /// A Collection of ICompressionProvider's that also allows them to be instantiated from an <see cref="IServiceProvider" />.
+    /// </summary>
+    public class CompressionProviderCollection : Collection<ICompressionProvider>
+    {
+        /// <summary>
+        /// Adds a type representing an <see cref="ICompressionProvider"/>.
+        /// </summary>
+        /// <remarks>
+        /// Provider instances will be created using an <see cref="IServiceProvider" />.
+        /// </remarks>
+        public void Add<TCompressionProvider>() where TCompressionProvider : ICompressionProvider
+        {
+            Add(typeof(TCompressionProvider));
+        }
+
+        /// <summary>
+        /// Adds a type representing an <see cref="ICompressionProvider"/>.
+        /// </summary>
+        /// <param name="providerType">Type representing an <see cref="ICompressionProvider"/>.</param>
+        /// <remarks>
+        /// Provider instances will be created using an <see cref="IServiceProvider" />.
+        /// </remarks>
+        public void Add(Type providerType)
+        {
+            if (providerType == null)
+            {
+                throw new ArgumentNullException(nameof(providerType));
+            }
+
+            if (!typeof(ICompressionProvider).IsAssignableFrom(providerType))
+            {
+                throw new ArgumentException($"The provider must implement {nameof(ICompressionProvider)}", nameof(providerType));
+            }
+
+            var factory = new CompressionProviderFactory(providerType);
+            Add(factory);
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.ResponseCompression/CompressionProviderFactory.cs
+++ b/src/Microsoft.AspNetCore.ResponseCompression/CompressionProviderFactory.cs
@@ -1,0 +1,48 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNetCore.ResponseCompression
+{
+    /// <summary>
+    /// This is a placeholder for the CompressionProviderCollection that allows creating the given type via
+    /// an <see cref="IServiceProvider" />.
+    /// </summary>
+    internal class CompressionProviderFactory : ICompressionProvider
+    {
+        internal CompressionProviderFactory(Type providerType)
+        {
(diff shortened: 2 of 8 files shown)


Tratcher:
@pranavkm

src/Microsoft.AspNetCore.ResponseCompression/CompressionProviderCollection.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// </summary>
+    public class CompressionProviderCollection : Collection<ICompressionProvider>
+    {
+        /// <summary>
+        /// Adds a type representing an <see cref="ICompressionProvider"/>.
+        /// </summary>
+        /// <remarks>
+        /// Provider instances will be created through dependency injection.
+        /// </remarks>
+        public void Add<T>()
```
pranavkm: `Add<TCompressionProvider>() where TCompressionProvider : ICompressionProvider, new()`?

src/Microsoft.AspNetCore.ResponseCompression/CompressionProviderFactory.cs
```diff
@@ -0,0 +1,48 @@
...
+            {
+                throw new ArgumentNullException(nameof(serviceProvider));
+            }
+
+            return (ICompressionProvider)ActivatorUtilities.CreateInstance(serviceProvider, ProviderType, Type.EmptyTypes);
+        }
+
+        string ICompressionProvider.EncodingName
+        {
+            get { throw new NotImplementedException(); }
```
pranavkm: `NotSupportedException`?

src/Microsoft.AspNetCore.ResponseCompression/CompressionProviderCollection.cs
```diff
@@ -0,0 +1,51 @@
...
+#if NETSTANDARD1_3
+using System.Reflection;
+#endif
+
+namespace Microsoft.AspNetCore.ResponseCompression
+{
+    /// <summary>
+    /// A Collection of ICompressionProvider's that also allows them to be instantiated from DI.
+    /// </summary>
+    public class CompressionProviderCollection : Collection<ICompressionProvider>
```
pranavkm: Wasn't part of the design, so I'm not super familiar. Would we want to consider a design like `FilterCollection` in Mvc since it looks pretty similar - https://github.com/aspnet/Mvc/blob/760c8f38678118734399c58c2dac981ea6e47046/src/Microsoft.AspNetCore.Mvc.Core/Filters/FilterCollection.cs?

Essentially,

* `ICompressionProviderMetadata`, 
* `ICompressionProvider : ICompressionProviderMetadata`
* `ICompressionProviderFactory : ICompressionProviderMetadata` 
* and this would be `Collection<ICompressionProviderMetadata>`

Obviously, filters are a much more useful extensibility point and I'm not sure this is.

src/Microsoft.AspNetCore.ResponseCompression/CompressionProviderCollection.cs
```diff
@@ -0,0 +1,51 @@
...
+using System;
+using System.Collections.ObjectModel;
+#if NETSTANDARD1_3
+using System.Reflection;
+#endif
+
+namespace Microsoft.AspNetCore.ResponseCompression
+{
+    /// <summary>
+    /// A Collection of ICompressionProvider's that also allows them to be instantiated from DI.
```
pranavkm: from ~~DI~~ an `<see cref="IServiceProvider" />.`

pranavkm: Couple of other places with the same comment.

src/Microsoft.AspNetCore.ResponseCompression/CompressionProviderCollection.cs
Tratcher: I can filter by ICompressionProvider, but `new()` is to restrictive, we want to inject parameters from DI.

pranavkm: Gotcha. For some reason I thought you were doing `Activator.CreateInstance` in the factory.

Tratcher:
Updated

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionProvider.cs
```diff
@@ -20,28 +20,37 @@ public class ResponseCompressionProvider : IResponseCompressionProvider
...
 
-            _providers = providers.ToArray();
+            _providers = options.Value.Providers.ToArray();
             if (_providers.Length == 0)
             {
-                _providers = new [] { new GzipCompressionProvider() };
+                // Use the factory so it can resolve IOptions<GzipCompressionProviderOptions> from DI.
+                _providers = new ICompressionProvider[] { new CompressionProviderFactory(typeof(GzipCompressionProvider)) };
+            }
+            for (int i = 0; i < _providers.Length; i++)
```
pranavkm: `var` all day every day.

src/Microsoft.AspNetCore.ResponseCompression/CompressionProviderCollection.cs
```diff
@@ -0,0 +1,51 @@
...
+        public void Add(Type providerType)
+        {
+            if (providerType == null)
+            {
+                throw new ArgumentNullException(nameof(providerType));
+            }
+
+            if (!typeof(ICompressionProvider).IsAssignableFrom(providerType))
+            {
+                throw new ArgumentException($"The provider must implement {nameof(ICompressionProvider)}", nameof(providerType));
```
Eilon: End message with a period.

src/Microsoft.AspNetCore.ResponseCompression/CompressionProviderFactory.cs
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Microsoft.AspNetCore.ResponseCompression
+{
+    /// <summary>
+    /// This is a placeholder for the CompressionProviderCollection that allows creating the given type via
+    /// an <see cref="IServiceProvider" />.
+    /// </summary>
+    internal class CompressionProviderFactory : ICompressionProvider
+    {
+        internal CompressionProviderFactory(Type providerType)
```
Eilon: It seems to me that logically the ctor, properties, and methods on this type should all be public, no? I realize the type is internal, but as far as member visibility goes the pattern we typically follow is, "what visibility would this have *if* the type was public"? And I think in this case they'd be public.

Issue #67, opened 2016-08-15
Url Rewrite Checklist


jkotalik:
**Features implemented in Url Rewrite:**

Elements/Attributes in rules:
- [x] Rules
- [x] Name
- [x] Enabled
- [x] PatternSyntax ECMAScript
- [x] ExactMatch
- [x] StopProcessing
- [x] Matches
- [x] Conditions: MatchAll
- [x] Conditions: MatchAny
- [x] TrackAllCaptures
- [x] IsFile
- [x] IsDirectory
- [x] IgnoreCase
- [x] Negate
- [x] Action:Rewrite
- [x] Action:Redirect
- [x] AppendQueryString
- [x] RedirectType
- [x] Action:AbortRequest

Server Variables (common ones supported for now, TODO on complete list)

**Features that are not implemented in UrlRewrite:**
- [x] Global Rules
- [ ] Outbound Rules
- [x] RewriteMaps
- [x] Custom Server Variables

Elements/Attributes in rules:
- [ ] WildCard
- [ ] ServerVariables 
- [ ] Action:CustomResponse
- [ ] LogRewrittenUrl

davidpeden3:
added:
- global rules support: https://github.com/aspnet/BasicMiddleware/pull/169
- rewrite map support: https://github.com/aspnet/BasicMiddleware/pull/168
- custom request/response headers via server variables: https://github.com/aspnet/BasicMiddleware/pull/184

jkotalik:
Closing due to supporting everything we need and we are not doing outbound rules.

PR #134, opened 2016-10-17
Added support for the AbortRequest action

mikaelm12:
[Background of the AbortRequest action](https://www.iis.net/learn/extensions/url-rewrite-module/url-rewrite-module-configuration-reference#AbortRequest_action)

> An AbortRequest action causes the URL Rewrite Module to drop the HTTP connection for the current request. The action does not have any parameters. Use of this action implies that no subsequent rules are evaluated for the current URL after this action is performed.

I also added an example to the sample. 
@Tratcher @natemcmaster

### DIFF
--- a/samples/RewriteSample/Startup.cs
+++ b/samples/RewriteSample/Startup.cs
@@ -16,7 +16,7 @@ namespace RewriteSample
             var options = new RewriteOptions()
                 .AddRedirect("(.*)/$", "$1")
                 .AddRewrite(@"app/(\d+)", "app?id=$1", skipRemainingRules: false)
-                .AddRedirectToHttps(302, 5001)
+                //.AddRedirectToHttps(302, 5001)
                 .AddIISUrlRewrite(env.ContentRootFileProvider, "UrlRewrite.xml")
                 .AddApacheModRewrite(env.ContentRootFileProvider, "Rewrite.txt");
 
--- a/samples/RewriteSample/Startup.cs
+++ b/samples/RewriteSample/Startup.cs
@@ -16,7 +16,7 @@ namespace RewriteSample
             var options = new RewriteOptions()
                 .AddRedirect("(.*)/$", "$1")
                 .AddRewrite(@"app/(\d+)", "app?id=$1", skipRemainingRules: false)
-                //.AddRedirectToHttps(302, 5001)
+                .AddRedirectToHttps(302, 5001)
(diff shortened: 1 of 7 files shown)


Tratcher:
Is there an issue tracking this feature work?

src/Microsoft.AspNetCore.Rewrite/Internal/UrlActions/AbortAction.cs
```diff
@@ -0,0 +1,16 @@
...
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Rewrite.Internal.UrlActions
+{
+    public class AbortAction : UrlAction
+    {
+        public override void ApplyAction(RewriteContext context, MatchResults ruleMatch, MatchResults condMatch)
+        {
+            context.HttpContext.Abort();
+            context.Result = RuleResult.EndResponse;
```
Tratcher: Log?

mikaelm12:
@Tratcher It's listed in the check list in https://github.com/aspnet/BasicMiddleware/issues/67 but no dedicated issue for just this feature. Should I open one?

Tratcher:
Yeah, there's few enough left on that list that we can open a bug for each (that we think are even possible) and triage them appropriately. There's already a bug for maps.

No fixing bugs that aren't assigned to the current milestone :-)

natemcmaster:
Implementation looks good to me. I'll let Chris do final sign off.

Tratcher:
@muratg for triage.

muratg:
Agreed with @Tratcher that we should file a bug for each remaining item and close #67. As of this one, this can wait till later. We'll triage once we ship preview1.

mikaelm12:
Rebased on dev. Can we check this in? @muratg  @Tratcher

samples/RewriteSample/Startup.cs
```diff
@@ -16,7 +16,7 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
             var options = new RewriteOptions()
                 .AddRedirect("(.*)/$", "$1")
                 .AddRewrite(@"app/(\d+)", "app?id=$1", skipRemainingRules: false)
-                .AddRedirectToHttps(302, 5001)
+                //.AddRedirectToHttps(302, 5001)
```
Tratcher: ?

muratg:
https://github.com/aspnet/BasicMiddleware/issues/136 triaged in 1.1.

Issue #139, opened 2016-10-18
Enabled rule attribute never used


BrennanConroy:
https://github.com/aspnet/BasicMiddleware/blob/dev/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs#L58

@mikaelm12 @muratg @jkotalik

jkotalik:
That issues doesn't fix the enable rule never being used. It should be added here:
https://github.com/aspnet/BasicMiddleware/blob/dev/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteRuleBuilder.cs#L31
as a check to see if we should be creating the rule in the first place.

mikaelm12:
The enabled flag on the builder is actually being used here to determine if each rule should be added
https://github.com/aspnet/BasicMiddleware/blob/d57149f9c04ff410f9a146ed686f15693e962137/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs#L51-L54

Issue #138, opened 2016-10-18
Parse invalid negate sets ignoreCase instead of negate in UrlRewrite


BrennanConroy:
https://github.com/aspnet/BasicMiddleware/blob/dev/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs#L152

@jkotalik @mikaelm12 @muratg

mikaelm12:
Fixed in https://github.com/aspnet/BasicMiddleware/pull/140

PR #140, opened 2016-10-19
Fixed Enable and negate flag issue. 

mikaelm12:
For issues https://github.com/aspnet/BasicMiddleware/issues/138 and https://github.com/aspnet/BasicMiddleware/issues/139. 

I also made the middleware throw when global rules are present since we don't actually handle them properly with the current implementation. 

@Tratcher @muratg @BrennanConroy

### DIFF
--- a/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
@@ -38,7 +38,7 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.IISUrlRewrite
                 return;
             }
 
-            if(string.Equals(rules.Name.ToString(), "GlobalRules", StringComparison.OrdinalIgnoreCase))
+            if (string.Equals(rules.Name.ToString(), "GlobalRules", StringComparison.OrdinalIgnoreCase))
             {
                 throw new NotSupportedException("Support for global rules has not been implemented yet");
             }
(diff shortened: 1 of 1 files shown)


BrennanConroy:
:shipit:

BrennanConroy:
Actually, this doesn't fix the enabled issue, see Justin's comment on the issue. Also, add a couple tests

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
```diff
@@ -38,6 +38,11 @@ private void ParseRules(XElement rules, IList<IISUrlRewriteRule> result)
                 return;
             }
 
+            if(string.Equals(rules.Name.ToString(), "GlobalRules", StringComparison.OrdinalIgnoreCase))
```
BrennanConroy: nit: space

mikaelm12:
It does fix the enabled flag issue. Justin was mistaken

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
```diff
@@ -59,6 +64,10 @@ private void ParseRuleAttributes(XElement rule, UrlRewriteRuleBuilder builder)
             {
                 builder.Enabled = true;
             }
+            else
+            {
+                builder.Enabled = enabled;
+            }
```
mikaelm12: That's a good idea. The only thing is that if we don't parse the rest of the rule we allow the user to enter invalid rules since the checking isn't happening. Not a huge deal since the rule won't be enabled bit still something to consider

mikaelm12:
Ping @Tratcher @BrennanConroy

BrennanConroy:
:shipit:

Issue #143, opened 2016-10-19
Fix NotSupportedException text in IISUrlRewrite Server Variables


BrennanConroy:
https://github.com/aspnet/BasicMiddleware/blob/dev/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/ServerVariables.cs

@mikaelm12

PR #150, opened 2016-10-21
Fixed exception messages for server variable parsing

mikaelm12:
Main thing here is that some of the exception messages for the server variable parsing was incorrect as pointed out in https://github.com/aspnet/BasicMiddleware/issues/143. I also did some refactoring to make the server variable parsing for IIS Url rewrite and Apache mod rewrite a little more similar.

Not urgent, just thought I'd open the pr.

@BrennanConroy @Tratcher @natemcmaster  @muratg

### DIFF
--- a/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/ServerVariables.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/ServerVariables.cs
@@ -9,15 +9,15 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.IISUrlRewrite
 {
     public static class ServerVariables
     {
-        public static PatternSegment FindServerVariable(string serverVariable)
+        public static PatternSegment FindServerVariable(string serverVariable, ParserContext context)
         {
             switch (serverVariable)
             {
                 // TODO Add all server variables here.
                 case "ALL_RAW":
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/ServerVariables.cs
```diff
@@ -55,7 +55,7 @@ public static PatternSegment FindServerVariable(string serverVariable)
                 case "REQUEST_URI":
                     return new UrlSegment();
                 default:
-                    throw new FormatException("Unrecognized server variable");
+                    throw new FormatException(Resources.FormatError_InputParserUnrecognizedParameter(serverVariable, context.Index));
```
natemcmaster: The code has become inconsistent on strings vs resources. We should add an issue to clean this up. (You don't need to add to this PR.)

mikaelm12:
@muratg Is this okay to merge to dev, or should we wait?

muratg:
@mikaelm12 yeah, it should be OK to merge to dev.

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/ServerVariables.cs
mikaelm12: Okay, I logged https://github.com/aspnet/BasicMiddleware/issues/153

Issue #148, opened 2016-10-20
Remove '?' prefix from QUERY_STRING server variable


BrennanConroy:
https://www.iis.net/learn/extensions/url-rewrite-module/url-rewrite-module-configuration-reference#Accessing_URL_Parts_from_a_Rewrite_Rule
Shows the value of `{QUERY_STRING}` from `http://www.mysite.com/content/default.aspx?tabid=2&subtabid=3` to be `tabid=2&subtabid=3`

I also, did a quick test with the IIS Rewrite Module and it doesn't seem to include `?` in the `{QUERY_STRING}`

jkotalik:
Good catch. Should be a one line fix in the server variables. Check mod_rewrite if you get the chance for its behavior.

BrennanConroy:
I will be checking mod_rewrite later

BrennanConroy:
mod_rewrite has the same issue since they both use the `QueryStringSegment` class

PR #158, opened 2016-10-24
Removed the ? prefix from the QUERY_STRING server variable

mikaelm12:
Addresses https://github.com/aspnet/BasicMiddleware/issues/148
@Tratcher @BrennanConroy

### DIFF
--- a/src/Microsoft.AspNetCore.Rewrite/Internal/PatternSegments/QueryStringSegment.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/Internal/PatternSegments/QueryStringSegment.cs
@@ -7,7 +7,7 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.PatternSegments
     {
         public override string Evaluate(RewriteContext context, MatchResults ruleMatch, MatchResults condMatch)
         {
-            return context.HttpContext.Request.QueryString.ToString();
+            return context.HttpContext.Request.QueryString.ToString().Substring(1);
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Rewrite/Internal/PatternSegments/QueryStringSegment.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/Internal/PatternSegments/QueryStringSegment.cs
@@ -7,7 +7,14 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.PatternSegments
     {
         public override string Evaluate(RewriteContext context, MatchResults ruleMatch, MatchResults condMatch)
         {
-            return context.HttpContext.Request.QueryString.ToString().Substring(1);
+            var queryString = context.HttpContext.Request.QueryString.ToString();
+
+            if (!string.IsNullOrEmpty(queryString))
+            {
+                return queryString.Substring(1);
+            }
+
+            return queryString;
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Rewrite/Internal/PatternSegments/QueryStringSegment.cs
```diff
@@ -7,7 +7,7 @@ public class QueryStringSegment : PatternSegment
     {
         public override string Evaluate(RewriteContext context, MatchResults ruleMatch, MatchResults condMatch)
         {
-            return context.HttpContext.Request.QueryString.ToString();
+            return context.HttpContext.Request.QueryString.ToString().Substring(1);
```
Tratcher: Throws ArgumentOutOfRangeException if there is no query

src/Microsoft.AspNetCore.Rewrite/Internal/PatternSegments/QueryStringSegment.cs
```diff
@@ -7,7 +7,14 @@ public class QueryStringSegment : PatternSegment
     {
         public override string Evaluate(RewriteContext context, MatchResults ruleMatch, MatchResults condMatch)
         {
-            return context.HttpContext.Request.QueryString.ToString();
+            var queryString = context.HttpContext.Request.QueryString.ToString();
```
jkotalik: Just to be sure queryString can never be null, right? If so you can probably use IsEmpty instead/

test/Microsoft.AspNetCore.Rewrite.Tests/PatternSegments/QueryStringSegmentTests.cs
```diff
@@ -12,16 +12,13 @@ public class QueryStringSegmentTests
         [Fact]
         public void QueryString_AssertSegmentIsCorrect()
         {
-            // Arrange
```
mikaelm12: Meh, they don't need to be there 😄

src/Microsoft.AspNetCore.Rewrite/Internal/PatternSegments/QueryStringSegment.cs
mikaelm12: There's actually no `string.IsEmpty` method in C#. It would have to be a string compare to the `string.Empty` field. I think I'll just leave null and empty check 😄

test/Microsoft.AspNetCore.Rewrite.Tests/IISUrlRewrite/ServerVariableTests.cs
```diff
@@ -22,7 +22,7 @@ public class ServerVariableTests
         [InlineData("HTTP_USER_AGENT", "useragent")]
         [InlineData("HTTP_CONNECTION", "connection")]
         [InlineData("HTTP_URL", "/foo")]
-        [InlineData("QUERY_STRING", "?bar=1")]
+        [InlineData("QUERY_STRING", "bar=1")]
```
Tratcher: Needs a test for an empty query

src/Microsoft.AspNetCore.Rewrite/Internal/PatternSegments/QueryStringSegment.cs
jkotalik: XD kinda surprised by that.

test/Microsoft.AspNetCore.Rewrite.Tests/IISUrlRewrite/ServerVariableTests.cs
jkotalik: How about null?

Issue #146, opened 2016-10-19
LogicalGrouping check is incorrect


BrennanConroy:
When parsing `logicalGrouping` in `Conditions` we use `RewriteTags.MatchType`, we need to add a `RewriteTags.LogicalGrouping` constant and use that for `logicalGrouping` parsing

@mikaelm12

mikaelm12:
Fixed in https://github.com/aspnet/BasicMiddleware/pull/160

Issue #145, opened 2016-10-19
patternSyntax parsing behavior


BrennanConroy:
Change `WildCard` to `Wildcard`
https://github.com/aspnet/BasicMiddleware/blob/dev/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/PatternSyntax.cs#L9
OR
Parse the Enum ignoring case
https://github.com/aspnet/BasicMiddleware/blob/dev/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs#L64

@mikaelm12

BrennanConroy:
Also look at the other Enums like MatchType, found examples of both uppercase and camelCase on the interwebz

mikaelm12:
We should definitely change the `WildCard` enum to `Wildcard` for consistency with the Url Rewrite reference doc and I think that we should also do case insensitive parsing of enums.

Thoughts @Tratcher @natemcmaster ?

natemcmaster:
> ... change the WildCard enum to Wildcard for consistency with the Url Rewrite reference doc and I think that we should also do case insensitive parsing of enums.

Yes to both.

mikaelm12:
Fixed in https://github.com/aspnet/BasicMiddleware/pull/160

PR #160, opened 2016-10-26
Fix enum parsing

mikaelm12:
Currently enum parsing isn't case sensitive and some of the enums have incorrect casing which is a bad combo..
Issue: https://github.com/aspnet/BasicMiddleware/issues/145
@BrennanConroy @natemcmaster @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/MatchType.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/MatchType.cs
@@ -5,8 +5,8 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.IISUrlRewrite
 {
     public enum MatchType
     {
-        pattern,
-        isFile,
-        isDirectory
+        Pattern,
+        IsFile,
+        IsDirectory
     }
 }
--- a/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/MatchType.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/MatchType.cs
@@ -5,7 +5,7 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.IISUrlRewrite
 {
     public enum MatchType
     {
-        Pattern,
+        pattern,
         isFile,
         isDirectory
     }
--- a/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/MatchType.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/MatchType.cs
@@ -6,7 +6,7 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.IISUrlRewrite
     public enum MatchType
     {
         Pattern,
-        IsFile,
-        IsDirectory
+        isFile,
+        isDirectory
     }
 }
--- a/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
@@ -223,7 +223,7 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.IISUrlRewrite
             }
 
             RedirectType redirectType;
-            if(urlAction.Attribute(RewriteTags.RedirectType) == null)
+            if (urlAction.Attribute(RewriteTags.RedirectType) == null)
             {
                 redirectType = RedirectType.Permanent;
             }
--- a/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
@@ -77,9 +77,13 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.IISUrlRewrite
             }
 
             PatternSyntax patternSyntax;
-            if (!Enum.TryParse(rule.Attribute(RewriteTags.PatternSyntax)?.Value, ignoreCase:true, result:out patternSyntax))
+            if (rule.Attribute(RewriteTags.PatternSyntax) == null)
             {
                 patternSyntax = PatternSyntax.ECMAScript;
+            }
+            else if (!Enum.TryParse(rule.Attribute(RewriteTags.PatternSyntax).Value, ignoreCase: true, result:out patternSyntax))
+            {
+                ThrowParameterFormatException(rule, "The patternSyntax parameter was not recognized");
             }
 
             bool stopProcessing;
@@ -135,9 +139,13 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.IISUrlRewrite
             }
 
             LogicalGrouping grouping;
-            if (!Enum.TryParse(conditions.Attribute(RewriteTags.MatchType)?.Value, ignoreCase:true, result:out grouping))
+            if (conditions.Attribute(RewriteTags.LogicalGrouping) == null)
             {
                 grouping = LogicalGrouping.MatchAll;
+            }
+            else if (!Enum.TryParse(conditions.Attribute(RewriteTags.LogicalGrouping).Value, ignoreCase: true, result: out grouping))
+            {
+                ThrowParameterFormatException(conditions, "The logicalGrouping parameter was not recognized");
             }
 
             bool trackingAllCaptures;
@@ -169,9 +177,13 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.IISUrlRewrite
             }
 
             MatchType matchType;
-            if (!Enum.TryParse(condition.Attribute(RewriteTags.MatchType)?.Value, ignoreCase:true, result:out matchType))
+            if (condition.Attribute(RewriteTags.MatchType) == null)
             {
                 matchType = MatchType.Pattern;
+            }
+            else if (!Enum.TryParse(condition.Attribute(RewriteTags.MatchType).Value, ignoreCase: true, result: out matchType))
+            {
+                ThrowParameterFormatException(condition, "The matchType parameter wasn't recognized");
             }
 
             var parsedInputString = condition.Attribute(RewriteTags.Input)?.Value;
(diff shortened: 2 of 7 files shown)


src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/MatchType.cs
```diff
@@ -5,8 +5,8 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.IISUrlRewrite
 {
     public enum MatchType
     {
-        Pattern,
-        IsFile,
-        IsDirectory
+        pattern,
```
davidfowl: This is anti C# casing rules. Why do you need to change the enum values if you're ignoring the case when parsing?

mikaelm12: Yeah, it didn't feel right writing it that way but I though that regardless of the parsing options it would be better for the enums to be consistent with the value casing from the Url Rewrite documentation

davidfowl: I think not. These should be reverted as they are never exposed to end user code. These are internal implementation details of our URLRewrite engine. So the casing should just be what it was before and we should parse case insensitive from the file format (which is the only thing that matters).

davidfowl:
You also need to add tests to verify the case insensitivity.

test/Microsoft.AspNetCore.Rewrite.Tests/IISUrlRewrite/MiddleWareTests.cs
```diff
@@ -336,5 +336,35 @@ public class MiddlewareTests
 
             Assert.Equal(response.Headers.Location.OriginalString, "/foo");
         }
+
+        [Theory]
+        [InlineData("IsFile")]
+        [InlineData("isfile")]
+        [InlineData("IsDirectory")]
+        [InlineData("isdirectory")]
+        public async Task VerifyIsFileAndIsDirectoryParsing(string matchType)
```
BrennanConroy: This isn't really testing that the parsing is working, before an invalid value would just make `matchType = Pattern` and now the same thing will still happen. We need to verify that `IsFile` and `IsDirectory` are being used

mikaelm12: The same thing wouldn't happen. If it was parsed incorrectly and the default value of `Pattern `was set for the `matchType` then it would throw because no pattern is set in the conditions that use `IsFile` and `IsDirectory`. It is being parsed correctly 😃

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
```diff
@@ -207,7 +207,7 @@ private void ParseUrlAction(XElement urlAction, UrlRewriteRuleBuilder builder, b
             }
 
             RedirectType redirectType;
-            if (!Enum.TryParse(urlAction.Attribute(RewriteTags.RedirectType)?.Value, out redirectType))
+            if (!Enum.TryParse(urlAction.Attribute(RewriteTags.RedirectType)?.Value, true, out redirectType))
```
mikaelm12: Do you think we should do this for all the enum parsing?

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
```diff
@@ -207,10 +207,17 @@ private void ParseUrlAction(XElement urlAction, UrlRewriteRuleBuilder builder, b
             }
 
             RedirectType redirectType;
-            if (!Enum.TryParse(urlAction.Attribute(RewriteTags.RedirectType)?.Value, true, out redirectType))
+            if(urlAction.Attribute(RewriteTags.RedirectType) == null)
             {
                 redirectType = RedirectType.Permanent;
             }
+            else
```
Tratcher: else if

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
```diff
@@ -207,10 +207,17 @@ private void ParseUrlAction(XElement urlAction, UrlRewriteRuleBuilder builder, b
...
+            {
+                if(!Enum.TryParse(urlAction.Attribute(RewriteTags.RedirectType)?.Value, ignoreCase: true, result: out redirectType))
```
Tratcher: the `?.` is no longer required, correct?

Tratcher: Better. Don't all the other enum parsers in this PR have the same issue with defaults?

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
```diff
@@ -246,5 +253,12 @@ private static void ThrowUrlFormatException(XElement element, string message, Ex
             var col = ((IXmlLineInfo)element).LinePosition;
             throw new FormatException(Resources.FormatError_UrlRewriteParseError(message, line, col), ex);
         }
+
+        private static void ThrowParameterFormatException(XElement element, string message)
+        {
+            var line = ((IXmlLineInfo)element).LineNumber;
+            var col = ((IXmlLineInfo)element).LinePosition;
+            throw new FormatException(Resources.FormatError_UrlRewriteParseError(message, line, col));
```
Tratcher: value?

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
Tratcher: check your spacing

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
mikaelm12: I was following the format of the rest of the exception handling in the file.

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
mikaelm12: > Don't all the other enum parsers in this PR have the same issue with defaults?

Yeah, I was going to fix the other scenarios in another PR but I'll just add them now

mikaelm12:
⬆️ 📅  @Tratcher @davidfowl 
I added the logicalGrouping rewrite tag  because the Enum parsing can't be done without it.

test/Microsoft.AspNetCore.Rewrite.Tests/IISUrlRewrite/FormatExceptionHandlingTests.cs
```diff
@@ -101,6 +101,62 @@ public class FormatExceptionHandlingTests
...
+            </conditions>
+            <action type=""Redirect"" url =""{R:1}"" />
+        </rule>
+    </rules>
+</rewrite>",
+            "Could not parse the UrlRewrite file. Message: 'The logicalGrouping parameter was not recognized'. Line number '5': '14'.")]
+        [InlineData(
+@"<rewrite>
+    <rules>
+        <rule name=""Remove trailing slash"" patternSyntax=""foo"">
```
Tratcher: if you do `patternSyntax=""""` does `value` come out `null` or `empty`?

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
mikaelm12: My b. I misunderstood your original comment. Got it

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
```diff
@@ -77,10 +77,14 @@ private void ParseRuleAttributes(XElement rule, UrlRewriteRuleBuilder builder)
             }
 
             PatternSyntax patternSyntax;
-            if (!Enum.TryParse(rule.Attribute(RewriteTags.PatternSyntax)?.Value, out patternSyntax))
+            if (rule.Attribute(RewriteTags.PatternSyntax) == null)
             {
                 patternSyntax = PatternSyntax.ECMAScript;
             }
+            else if (!Enum.TryParse(rule.Attribute(RewriteTags.PatternSyntax).Value, ignoreCase: true, result:out patternSyntax))
```
davidfowl: nit: format document

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
```diff
@@ -246,5 +266,12 @@ private static void ThrowUrlFormatException(XElement element, string message, Ex
             var col = ((IXmlLineInfo)element).LinePosition;
             throw new FormatException(Resources.FormatError_UrlRewriteParseError(message, line, col), ex);
         }
+
+        private static void ThrowParameterFormatException(XElement element, string message)
+        {
+            var line = ((IXmlLineInfo)element).LineNumber;
```
davidfowl: nit: cast once.

test/Microsoft.AspNetCore.Rewrite.Tests/IISUrlRewrite/FormatExceptionHandlingTests.cs
mikaelm12: Empty and it throws

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
```diff
@@ -81,9 +81,9 @@ private void ParseRuleAttributes(XElement rule, UrlRewriteRuleBuilder builder)
             {
                 patternSyntax = PatternSyntax.ECMAScript;
             }
-            else if (!Enum.TryParse(rule.Attribute(RewriteTags.PatternSyntax).Value, ignoreCase: true, result:out patternSyntax))
+            else if (!Enum.TryParse(rule.Attribute(RewriteTags.PatternSyntax).Value, ignoreCase: true, result: out patternSyntax))
             {
-                ThrowParameterFormatException(rule, "The patternSyntax parameter was not recognized");
+                ThrowParameterFormatException(rule, $"The {RewriteTags.PatternSyntax} parameter: {rule.Attribute(RewriteTags.PatternSyntax).Value}, was not recognized");
```
Tratcher: `parameter '{rule.Attribute(RewriteTags.PatternSyntax).Value}' was not recognized

jkotalik:
Besides any grammar/exception stuff, :ship: :it:

mikaelm12:
Also fixes issue https://github.com/aspnet/BasicMiddleware/issues/146

Issue #152, opened 2016-10-21
mod_rewrite redirect flag should allow no status code value


BrennanConroy:
On a `RewriteRule` the `[R]` flag allows an optional status code value but currently we require a status code value

jkotalik:
Default to 302?

BrennanConroy:
Looks like it
http://httpd.apache.org/docs/current/rewrite/flags.html#flag_r

Tratcher:
hmm, what about these other parts?
`The status code specified need not necessarily be a redirect (3xx) status code. However, if a status code is outside the redirect range (300-399) then the substitution string is dropped entirely, and rewriting is stopped as if the L were used.`

`In addition to response status codes, you may also specify redirect status using their symbolic names: temp (default), permanent, or seeother.`

PR #163, opened 2016-11-02
Added default status code for mod_rewrite redirect

mikaelm12:
Currently the middleware throws when there is no status code with the redirect flag. It should default to 302.
Addresses https://github.com/aspnet/BasicMiddleware/issues/152
@BrennanConroy @Tratcher @natemcmaster

### DIFF
--- a/src/Microsoft.AspNetCore.Rewrite/Internal/ApacheModRewrite/FlagParser.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/Internal/ApacheModRewrite/FlagParser.cs
@@ -75,17 +75,21 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.ApacheModRewrite
             var flags = new Flags();
             foreach (var token in tokens)
             {
-                var hasPayload = token.Split('=');
+                var payload = token.Split('=');
 
                 FlagType flag;
-                if (!_ruleFlagLookup.TryGetValue(hasPayload[0], out flag))
+                if (!_ruleFlagLookup.TryGetValue(payload[0], out flag))
                 {
-                    throw new FormatException($"Unrecognized flag: '{hasPayload[0]}'");
+                    throw new FormatException($"Unrecognized flag: '{payload[0]}'");
                 }
 
-                if (hasPayload.Length == 2)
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Rewrite/Internal/ApacheModRewrite/FlagParser.cs
```diff
@@ -75,17 +75,21 @@ public Flags Parse(string flagString)
...
 
-                if (hasPayload.Length == 2)
+                if (payload.Length == 2)
                 {
-                    flags.SetFlag(flag, hasPayload[1]);
+                    flags.SetFlag(flag, payload[1]);
+                }
+                else if (string.Equals(payload[0], "R", StringComparison.OrdinalIgnoreCase))
+                {
+                    flags.SetFlag(flag, "302");
```
Tratcher: The raw string parser seems too low level to be applying this kind of default. Where is the "R" flag interpreted? Where was the exception being thrown?

Tratcher: I think the fix belongs here:
https://github.com/aspnet/BasicMiddleware/blob/1db5a9e58f5b40d316af7a82d92022fa35fe3cc0/src/Microsoft.AspNetCore.Rewrite/Internal/ApacheModRewrite/RuleBuilder.cs#L201

mikaelm12: Yeah, I had the exact same conversation with @BrennanConroy about this earlier today. Setting the default status code make writing the tests more complicated, but the implementation makes more sense. I'll change it

BrennanConroy: Why you not listen to me?!

BrennanConroy:
Also, @Tratcher mentioned in the issue that there are a couple values you can use instead of status codes

mikaelm12:
Closing this. Opened https://github.com/aspnet/BasicMiddleware/pull/164

PR #167, opened 2016-11-04
Created public API baselines

javiercn:


javiercn:
@Tratcher I believe these are the only ones that are 1.0.0

Issue #59, opened 2016-08-05
Url Rewrite - No distinction between global and regular rules.


jkotalik:
Currently global rules are treated like normal rules, while rather they should be treated to operate on both the path and path base (AFAIK).
Global rules:

> <globalRules> – Rules in this collection can be defined only on the server level. Global rules are used to define server-wide URL rewriting logic. These rules are defined within the ApplicationHost.config file, and they cannot be overridden or disabled on any lower configuration levels. Global rules always operate on the absolute URL's path (that is, the requested URI without the server name). These rules are evaluated early in the IIS request-processing pipeline (PreBeginRequest event).

davidpeden3:
> Wait, what requires forwarding? global rules? I don't see that in the description, only that it mutates the full url.

@Tratcher you may be technically correct (i'm not sure, to be honest) and that's what i have implemented. for my use-case, we are using global rules specifically to create a reverse proxy. i went ahead and created a reverse proxy middleware which works for single requests but i'm concerned about concurrency and scale.

```csharp
public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory msLoggerFactory, ILogger logger)
{
	msLoggerFactory.AddSerilog(logger);
	RewriteOptions options = new RewriteOptions().AddIISUrlRewrite(env.ContentRootFileProvider, "RewriteRules.xml");
	app.UseRewriter(options);
	app.RunReverseProxy();
}
```

without adding `app.RunReverseProxy();`, the server returns a `404` which makes perfect sense. we require the request forwarding.

i'm not really sure what the purpose of a global rule that rewrites the full url would be if it were not being used to implement a reverse proxy but i guess that's technically out of scope for this issue. it seems like a reverse proxy middleware (as i have implemented separately as seen above) would be a great additional sub-project for this repo. definitely interested to hear your thoughts.

i'll try to get the PRs for both issues (this one and https://github.com/aspnet/BasicMiddleware/issues/55) submitted by this weekend, if not sooner. still need to write some unit tests and do a `git pull`. you guys appear to have changed some of the code since i originally cloned the repo and fixed one of the bugs that i discovered and also fixed while doing this work.

mikaelm12:
Resolved in https://github.com/aspnet/BasicMiddleware/pull/169

PR #169, opened 2016-11-04
#59 add IIS global rules support with tests

david-peden-q2:
i did not add a test for the actual middleware execution. there currently aren't any and i didn't want to introduce any sort of additional dependencies (mock frameworks, integration tests, etc.) as a part of this PR. as the middleware is written, it's somewhat difficult to get at the part of the code that was modified. i suspect that you guys will have feedback on this part anyway. ;)


Tratcher:
Thanks for the PRs. FYI we're wrapping up a milestone right now so we'll come back to this for the next milestone.

test/Microsoft.AspNetCore.Rewrite.Tests/IISUrlRewrite/FileParserTests.cs
```diff
@@ -128,6 +128,37 @@ public void RuleParse_ParseMultipleRules()
...
+                                    <action type=""Rewrite"" url =""article.aspx?id={R:1}&amp;title={R:2}"" />
+                                </rule>
+                            </rules>
+                        </rewrite>";
+
+            // act
+            var rules = new UrlRewriteFileParser().Parse(new StringReader(xml));
+
+            // assert
+            Assert.Equal(2, rules.Count);
```
mikaelm12: You should use the `AssertUrlRewriteRuleEquality` like the other tests in this file.

davidpeden3: @mikaelm12 frankly, i don't agree. that's an awful lot of setup for no real value. it also distracts from what the test is actually testing. beyond that, it would require that the existing method be updated to be aware of global rules and then all other tests that utilize that method would have to be changed as well. if you still think it should, i would appreciate seeing an example of how this would look.

mikaelm12: So yeah, I agree that I don't really think its worth it to change the rule equality method to be aware of global rules. I'd say check the global attribute the way its currently done and in addition ensure equality a bit more thoroughly with the `AssertUrlRewriteRuleEquality` method. It also would be nice to stay somewhat consistent with the other tests here.
Thoughts @Tratcher ?

src/Microsoft.AspNetCore.Rewrite/RewriteMiddleware.cs
```diff
@@ -75,25 +76,13 @@ public Task Invoke(HttpContext context)
...
-                        uriBuilder.Path = context.Request.Path;
-                        if (context.Request.QueryString.HasValue)
-                        {
-                            uriBuilder.Query = context.Request.QueryString.Value;
-                        }
-                        return uriBuilder.Uri.AbsoluteUri;
-                    }
-                    return context.Request.Path + context.Request.QueryString;
-                });
+                    rewriteContext.GlobalRule = iisRule.Global;
```
jkotalik: See comment above, I feel like during parsing, we can set if an IIS rule is a global rule or not.

src/Microsoft.AspNetCore.Rewrite/Extensions/HttpRequestExtensions.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Rewrite.Extensions
+{
+    public static class HttpRequestExtensions
+    {
+        public static Uri ToUri(this HttpRequest request)
```
jkotalik: @Tratcher @david-peden-q2 This should be in HttpAbstractions?

Tratcher: No, we do not use System.Uri, it makes too many modifications on its own. Remove this and use https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http.Extensions/UriHelper.cs

davidpeden3: cool, didn't know about `UriHelper`. i changed the implementation to use it. @Tratcher are you saying get rid of the extension method altogether? it's a lot of set up at each call site if so.

test/Microsoft.AspNetCore.Rewrite.Tests/IISUrlRewrite/FileParserTests.cs
davidpeden3: not trying to be difficult but just want to point out that the point of the test is to assert that file parsing works correctly with regard to identifying global vs local rules. doing the test setup to assert rules are fully parsed correctly distracts from the tests goal and is also redundant with other tests that specifically assert other aspects of rule parsing. if you insist, i'll make the suggested change but it doesn't feel right to me.

src/Microsoft.AspNetCore.Rewrite/Internal/PatternSegments/UrlSegment.cs
```diff
@@ -1,13 +1,15 @@
...
+using Microsoft.AspNetCore.Rewrite.Extensions;
+
 namespace Microsoft.AspNetCore.Rewrite.Internal.PatternSegments
 {
     public class UrlSegment : PatternSegment
     {
         public override string Evaluate(RewriteContext context, MatchResults ruleMatch, MatchResults condMatch)
         {
-            return context.HttpContext.Request.Path;
+            return context.GlobalRule ? context.HttpContext.Request.ToUri().AbsoluteUri : (string)context.HttpContext.Request.Path;
```
davidpeden3: i agree with you and that's what i started to do. `IISUrlRewriteRule` does have the property set but the issue i ran into is that inside of `UrlSegment`, doesn't have visibility into the rule. i thought about pushing the property in explicitly but that would've made a sweeping mess. i figured it was easiest/best given the current design to piggyback on the context as it required minimal changes. happy to discuss other ideas.

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/IISUrlRewriteRule.cs
```diff
@@ -13,16 +13,19 @@ public class IISUrlRewriteRule : IRule
...
         public IList<Condition> Conditions { get; }
         public UrlAction Action { get; }
+        public bool Global { get; }
 
         public IISUrlRewriteRule(string name,
             UrlMatch initialMatch,
             IList<Condition> conditions,
-            UrlAction action)
+            UrlAction action,
+            bool? global = null)
```
IEvangelist: Instead of `bool? global = null`, why not `bool global = false`? The  `.ctor` is falling back to `false` anyways. Also, why not `IsGlobal` instead?

davidpeden3: they are not equivalent. your suggestion would require that all pre-existing methods be updated to pass an explicit value. that may be desirable but i was going for least surface area change. regarding `IsGlobal` vs `Global`, i happen to share your preference. having said that, i was following the lead established elsewhere: https://github.com/aspnet/BasicMiddleware/blob/623333a919c34e63ae98a276564e03f31f9cb1cf/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteRuleBuilder.cs#L18.

IEvangelist: @davidpeden3 based on what you currently have they are in fact the same, perhaps I should have clarified this a little further. Consider the following:
```
public class Foo
{
    public bool Bar { get; }
    public Foo(bool bar = false) // My proposal
    {
        Bar = bar;
    }
}
```
Is cleaner as a tri-state `bool` is meaningless as `null` is always `false`.
```
public class Foo
{
    public bool Bar { get; }
    public Foo(bool? bar = null) // What you currently have
    {
        Bar = bar ?? false;
    }
}
```
Either way you can instantiate `Foo` like this:
```
var foo = new Foo(); // foo.Bar is false
var foo = new Foo(false); // foo.Bar is false
var foo = new Foo(true); // foo.Bar is true
```

davidpeden3: hmm... yeah, not sure what i was thinking. o.O

fixed. ;)

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
```diff
@@ -22,35 +22,28 @@ public IList<IISUrlRewriteRule> Parse(TextReader reader)
             if (xmlRoot != null)
             {
                 var result = new List<IISUrlRewriteRule>();
-                // TODO Global rules are currently not treated differently than normal rules, fix.
-                // See: https://github.com/aspnet/BasicMiddleware/issues/59
-                ParseRules(xmlRoot.Descendants(RewriteTags.GlobalRules).FirstOrDefault(), result);
-                ParseRules(xmlRoot.Descendants(RewriteTags.Rules).FirstOrDefault(), result);
+                ParseRules(xmlRoot.Descendants(RewriteTags.GlobalRules).FirstOrDefault(), result, true);
```
Tratcher: use named parameters for constants

src/Microsoft.AspNetCore.Rewrite/Internal/PatternSegments/UrlSegment.cs
```diff
@@ -1,13 +1,15 @@
...
+using Microsoft.AspNetCore.Http.Extensions;
+
 namespace Microsoft.AspNetCore.Rewrite.Internal.PatternSegments
 {
     public class UrlSegment : PatternSegment
     {
         public override string Evaluate(RewriteContext context, MatchResults ruleMatch, MatchResults condMatch)
         {
-            return context.HttpContext.Request.Path;
+            return context.GlobalRule ? context.HttpContext.Request.GetEncodedUrl() : (string)context.HttpContext.Request.Path;
```
Tratcher: @mikaelm12 We need to check if this gives the expected escaping compared with the native module.

src/Microsoft.AspNetCore.Rewrite/RewriteMiddleware.cs
```diff
@@ -74,8 +76,13 @@ public Task Invoke(HttpContext context)
 
             foreach (var rule in _options.Rules)
             {
+                var iisRule = rule as IISUrlRewriteRule;
```
Tratcher: No, we shouldn't have to special case any rule types at this level. This should be contained within ApplyRule.

src/Microsoft.AspNetCore.Rewrite/RewriteMiddleware.cs
```diff
@@ -74,8 +76,13 @@ public Task Invoke(HttpContext context)
...
+                if (iisRule != null)
+                {
+                    rewriteContext.GlobalRule = iisRule.Global;
+                }
                 rule.ApplyRule(rewriteContext);
-                var currentUrl = new Lazy<string>(() => context.Request.Path + context.Request.QueryString);
+                var currentUrl = new Lazy<string>(() => context.Request.GetEncodedUrl());
```
Tratcher: This lazy is pointless, remove it and inline the variable.

src/Microsoft.AspNetCore.Rewrite/RewriteContext.cs
```diff
@@ -34,6 +34,11 @@ public class RewriteContext
         /// </summary>
         public RuleResult Result { get; set; }
 
+        /// <summary>
+        /// Represents whether or not the rule is a global rule
+        /// </summary>
+        public bool GlobalRule { get; set; }
```
Tratcher: Let's try to remove this as it's specific to only one rule system. it's only used in UrlSegment? Maybe we need a separate GlobalRuleUrlSegment instead.

test/Microsoft.AspNetCore.Rewrite.Tests/PatternSegments/GlobalRuleUrlSegmentTests.cs
```diff
@@ -0,0 +1,36 @@
...
+using Microsoft.AspNetCore.Rewrite.Internal.PatternSegments;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Rewrite.Tests.PatternSegments
+{
+    public class GlobalRuleUrlSegmentTests
+    {
+        [Theory]
+        [InlineData("http", "localhost", 80, "", "http://localhost:80/")]
+        [InlineData("http", "localhost", 80, "/foo/bar", "http://localhost:80/foo/bar")]
```
Tratcher: We need to check something that has escaped characters in the path, like %3A. Also add test for the query.

test/Microsoft.AspNetCore.Rewrite.Tests/PatternSegments/UrlSegmentTests.cs
```diff
@@ -9,18 +9,25 @@ namespace Microsoft.AspNetCore.Rewrite.Tests.PatternSegments
 {
     public class UrlSegmentTests
     {
-        [Fact]
-        public void LocalPortSegment_AssertSegmentIsCorrect()
+        [Theory]
+        [InlineData("http", "localhost", 80, "/foo/bar", "/foo/bar")]
```
Tratcher: Include something that would be un-escaped like `:`

test/Microsoft.AspNetCore.Rewrite.Tests/PatternSegments/GlobalRuleUrlSegmentTests.cs
Tratcher: And go back and compare the escaping behavior to IIS

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/ServerVariables.cs
```diff
@@ -9,7 +9,15 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.IISUrlRewrite
 {
     public static class ServerVariables
     {
-        public static PatternSegment FindServerVariable(string serverVariable, ParserContext context)
+		/// <summary>
```
Tratcher: Did the formatting get messed up here? It shows up strange in the diff.

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/InputParser.cs
```diff
@@ -13,24 +13,25 @@ public class InputParser
...
-        /// <returns>A new <see cref="Pattern"/>, containing a list of <see cref="PatternSegment"/></returns>
-        public Pattern ParseInputString(string testString)
+	    /// <summary>
+	    /// Creates a pattern, which is a template to create a new test string to
+	    /// compare to the condition. Can contain server variables, back references, etc.
+	    /// </summary>
+	    /// <param name="testString"></param>
+	    /// <param name="global"></param>
+	    /// <returns>A new <see cref="Pattern"/>, containing a list of <see cref="PatternSegment"/></returns>
+	    public Pattern ParseInputString(string testString, bool global)
```
Tratcher: indentation?

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/IISUrlRewriteRule.cs
```diff
@@ -14,18 +14,21 @@ public class IISUrlRewriteRule : IRule
...
         public bool TrackAllCaptures { get; }
+        public bool Global { get; }
 
         public IISUrlRewriteRule(string name,
             UrlMatch initialMatch,
             IList<Condition> conditions,
             UrlAction action,
-            bool trackAllCaptures)
+            bool trackAllCaptures,
+            bool global = false)
```
Tratcher: remove the default parameter. Add a second constructor if you need to.

Issue #141, opened 2016-10-19
Add trackAllCaptures for Back References in Url Rewrite


BrennanConroy:
e.g. 

```
<condtions trackAllCaptures="true">
  <add input="{QUERY_STRING}" pattern="id=(20)" />
  <add input="{HTTP_HOST}" pattern="(.*)" />
</conditions>
<action type="Redirect" url="https://{C:2}" />
```

@mikaelm12

BrennanConroy:
Remember, trackAllCaptures is part of the back reference feature
https://github.com/aspnet/BasicMiddleware/blob/89a6f453ce5eb85a0c41cf3153d8f790a5fa2103/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs#L133

BrennanConroy:
The correct name is `trackAllCaptures` not `trackingAllCaptures`
https://github.com/aspnet/BasicMiddleware/blob/89a6f453ce5eb85a0c41cf3153d8f790a5fa2103/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/RewriteTags.cs#L24

BrennanConroy:
`BackReference` uses `RegularExpressions.GroupCollection` which doesn't have additive operators, so we'll have to change the structure used to contain the back references.

muratg:
@mikaelm12 Let's fix the enum name in 1.1. And investigate the back referencing issue. If the potential fix for that is not breaking, we can do it in 1.2 instead.

mikaelm12:
Okay. I changed the enum in https://github.com/aspnet/BasicMiddleware/pull/166
This bug should be moved to 1.2 though

mikaelm12:
Added in https://github.com/aspnet/BasicMiddleware/pull/178

PR #178, opened 2016-11-24
Adding support for trackAllCaptures

Tratcher:
#141 @mikaelm12 @BrennanConroy 
Does this make more sense? Still has a few rough edges...

### DIFF
--- a/src/Microsoft.AspNetCore.Rewrite/Internal/BackReferenceCollection.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/Internal/BackReferenceCollection.cs
@@ -31,7 +31,7 @@ namespace Microsoft.AspNetCore.Rewrite.Internal
         {
             get
             {
-                if(index < _backReferences.Count)
+                if (index < _backReferences.Count)
                 {
                     return _backReferences[index];
                 }
--- a/src/Microsoft.AspNetCore.Rewrite/Internal/BackReferenceCollection.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/Internal/BackReferenceCollection.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.Text.RegularExpressions;
 
@@ -8,7 +9,7 @@ namespace Microsoft.AspNetCore.Rewrite.Internal
 {
     public class BackReferenceCollection
     {
-        private List<string> backReferences = new List<string>();
+        private List<string> _backReferences = new List<string>();
 
         public BackReferenceCollection(GroupCollection references)
         {
@@ -16,27 +17,36 @@ namespace Microsoft.AspNetCore.Rewrite.Internal
             {
                 for (var i = 0; i < references.Count; i++)
                 {
-                    backReferences.Add(references[i].Value);
+                    _backReferences.Add(references[i].Value);
                 }
             }
         }
 
         public BackReferenceCollection(string reference)
         {
-            backReferences.Add(reference);
+            _backReferences.Add(reference);
         }
 
         public string this[int index]
         {
-            // TODO: index out of range error handling?
-            get { return backReferences[index]; }
+            get
+            {
+                if(index < _backReferences.Count)
+                {
+                    return _backReferences[index];
+                }
+                else
+                {
+                    throw new IndexOutOfRangeException($"Cannot access back reference at index {index}. Only {_backReferences.Count} back references were captured.");
+                }
+            }
         }
 
         public void Add(BackReferenceCollection references)
         {
             if (references != null)
             {
-                backReferences.AddRange(references.backReferences);
+                _backReferences.AddRange(references._backReferences);
             }
         }
     }
--- a/src/Microsoft.AspNetCore.Rewrite/Internal/BackReferenceCollection.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/Internal/BackReferenceCollection.cs
@@ -1,0 +1,43 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.AspNetCore.Rewrite.Internal
+{
+    public class BackReferenceCollection
+    {
+        private List<string> backReferences = new List<string>();
+
+        public BackReferenceCollection(GroupCollection references)
+        {
+            if (references != null)
+            {
+                for (var i = 0; i < references.Count; i++)
+                {
+                    backReferences.Add(references[i].Value);
+                }
+            }
+        }
+
+        public BackReferenceCollection(string reference)
+        {
+            backReferences.Add(reference);
+        }
+
+        public string this[int index]
+        {
+            // TODO: index out of range error handling?
+            get { return backReferences[index]; }
+        }
+
+        public void Add(BackReferenceCollection references)
+        {
+            if (references != null)
+            {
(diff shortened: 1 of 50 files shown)


src/Microsoft.AspNetCore.Rewrite/Internal/ConditionEvaluator.cs
```diff
@@ -24,18 +24,26 @@ public static MatchResults Evaluate(IEnumerable<Condition> conditions, RewriteCo
...
+                else if (!condResult.Success)
                 {
-                    return prevCond;
+                    return condResult;
                 }
+
+                if (condResult.Success && trackAllCaptures && prevResult?.BackReferences != null)
+                {
+                    prevResult.BackReferences.Add(condResult.BackReferences);
+                    condResult.BackReferences = prevResult.BackReferences;
```
Tratcher: Hmm, this may end up setting MatchResult.EmptySuccess. Should change this so the loop tracks backrefrences independent of MatchResult. In fact, there's no reason to hold on to the prior match result, only the backrefrences. Change Evaluate to take in references rather than results.

Tratcher:
@mikaelm12 go ahead and take over this PR/branch.

src/Microsoft.AspNetCore.Rewrite/Internal/BackReferenceCollection.cs
```diff
@@ -0,0 +1,43 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.AspNetCore.Rewrite.Internal
+{
+    public class BackReferenceCollection
+    {
+        private List<string> backReferences = new List<string>();
```
jkotalik: May want lazy initialization here. Also IList or enumerable may be preferred here, but would need to remove AddRange.

src/Microsoft.AspNetCore.Rewrite/Internal/BackReferenceCollection.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        public BackReferenceCollection(string reference)
+        {
+            backReferences.Add(reference);
+        }
+
+        public string this[int index]
+        {
+            // TODO: index out of range error handling?
+            get { return backReferences[index]; }
```
jkotalik: Yeah definitely handle that gracefully

src/Microsoft.AspNetCore.Rewrite/Internal/BackReferenceCollection.cs
davidfowl: nit: style where is the underscore...

mikaelm12: Not sure lazy initialization of backReferences would help here.

jkotalik: GroupCollection has a Count property, you can check if Count == 0 for backreferences.

jkotalik: Oh shoot, wont it always contain at least one element for the whole string? Then you are probably right

mikaelm12: Yeah, exactly. Every time its created we immediately add something to the backReferences list so using lazy initialization at least to my knowledge wouldn't really help.

src/Microsoft.AspNetCore.Rewrite/Internal/BackReferenceCollection.cs
```diff
@@ -1,42 +1,52 @@
...
+            _backReferences.Add(reference);
         }
 
         public string this[int index]
         {
-            // TODO: index out of range error handling?
-            get { return backReferences[index]; }
+            get
+            {
+                if(index < _backReferences.Count)
```
BrennanConroy: nit: space

mikaelm12:
Ping @BrennanConroy @Tratcher 😃

BrennanConroy:
🚢 :it:

src/Microsoft.AspNetCore.Rewrite/Internal/ConditionEvaluator.cs
Tratcher: Ping

mikaelm12:
@muratg

src/Microsoft.AspNetCore.Rewrite/Internal/ConditionEvaluator.cs
Tratcher: Still need to re-do this logic so that we track the backreference collections directly rather than updating MatchResults.

src/Microsoft.AspNetCore.Rewrite/Internal/BackReferenceCollection.cs
```diff
@@ -0,0 +1,53 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.AspNetCore.Rewrite.Internal
+{
+    public class BackReferenceCollection
+    {
+        private List<string> _backReferences = new List<string>();
```
muratg: nit: var

src/Microsoft.AspNetCore.Rewrite/Internal/ConditionEvaluator.cs
```diff
@@ -8,9 +8,9 @@ namespace Microsoft.AspNetCore.Rewrite.Internal
     public static class ConditionHelper
     {
 
-        public static MatchResults Evaluate(IEnumerable<Condition> conditions, RewriteContext context, MatchResults ruleMatch)
+        public static MatchResults Evaluate(IEnumerable<Condition> conditions, RewriteContext context, MatchResults ruleMatch, bool trackAllCaptures = false)
```
muratg: Yeah, I'd recommend an override which passes `false` instead.

src/Microsoft.AspNetCore.Rewrite/Internal/BackReferenceCollection.cs
moozzyk: ?

muratg: *ignore* lol

src/Microsoft.AspNetCore.Rewrite/Internal/ConditionEvaluator.cs
mikaelm12: Wouldn't that break the `isFile`  and `isDirectory` use case where we can have success but no BackReferences?

Tratcher: You still need to track success, but it can also be stored in a local variable, you don't need to flow the whole MatchResult.

src/Microsoft.AspNetCore.Rewrite/Internal/ConditionEvaluator.cs
```diff
@@ -7,10 +7,14 @@ namespace Microsoft.AspNetCore.Rewrite.Internal
 {
     public static class ConditionHelper
     {
-
-        public static MatchResults Evaluate(IEnumerable<Condition> conditions, RewriteContext context, MatchResults ruleMatch)
+        public static MatchResults Evaluate(IEnumerable<Condition> conditions, RewriteContext context, BackReferenceCollection backReferences)
+        {
+            return Evaluate(conditions, context, backReferences, trackAllCaptures: false);
+        }
+        public static MatchResults Evaluate(IEnumerable<Condition> conditions, RewriteContext context, BackReferenceCollection backReferences, bool trackAllCaptures)
```
Tratcher: nit missing an empty line

muratg:
:shipit:

PR #181, opened 2016-12-02
Added more GZIP compressable MIME types

RehanSaeed:
The MIME type list for what can be compressed is a bit basic. There are several other formats which can be compressed, including images and font file formats.

### DIFF
--- a/src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionDefaults.cs
+++ b/src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionDefaults.cs
@@ -18,15 +18,22 @@ namespace Microsoft.AspNetCore.ResponseCompression
         {
             // General
             "text/plain",
-            // Static files
             "text/css",
             "application/javascript",
-            // MVC
             "text/html",
             "application/xml",
             "text/xml",
+            "application/rss+xml",
+            "application/atom+xml",
             "application/json",
-            "text/json",
+            "text/json"
+            // Images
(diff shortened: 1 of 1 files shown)


Tratcher:
Please open an issue with the recommended types and justifications for each. This was intentionally not an exhaustive list to start with. Everybody is going to have their own types they want and we don't want to maintain a huge list, only types that more than 50-80% of apps will use.

What did you think of the API for adding your own types in Startup?
https://github.com/aspnet/BasicMiddleware/blob/dev/samples/ResponseCompressionSample/Startup.cs#L28

davidfowl:
> What did you think of the API for adding your own types in Startup?
https://github.com/aspnet/BasicMiddleware/blob/dev/samples/ResponseCompressionSample/Startup.cs#L28

That API is pretty awful. Especially because of the lack of append... Though even with append it's strange to use `IEnumerable<T>` that way. Why isn't the list mutable (like MVC's model binders)

RehanSaeed:
Agreed that a mutable collection would have been more natural, it's what I instinctively looked for but looking at the source found it to be an array. I've raised this https://github.com/aspnet/BasicMiddleware/issues/182 issue with my justifications for each MIME type.

Tratcher:
Mutable collections can't be added to in inline constructors.

muratg:
@glennc will be considering the work required and will use https://github.com/aspnet/BasicMiddleware/issues/182 to track the work.

Closing this one. Thanks @RehanSaeed!

Issue #183, opened 2016-12-05
Custom request/response headers in IIS via server variables


davidpeden3:
https://www.iis.net/learn/extensions/url-rewrite-module/url-rewrite-module-20-configuration-reference#Note_about_request_headers_

davidpeden3:
PR: https://github.com/aspnet/BasicMiddleware/pull/184

PR #184, opened 2016-12-05
#183 add support for custom request/response headers via server variables

davidpeden3:


jkotalik:
Thanks for all the contributions; I'll get to looking at these later today 🐐

src/Microsoft.AspNetCore.Rewrite/Extensions/HttpRequestExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+    public static class HttpRequestExtensions
+    {
+        public static Uri ToUri(this HttpRequest request)
+        {
+            var uriBuilder = new UriBuilder(request.Scheme, request.Host.Host);
+            if (request.Host.Port.HasValue)
+            {
+                uriBuilder.Port = request.Host.Port.Value;
+            }
+            uriBuilder.Path = request.Path;
```
Tratcher: Missing PathBase

src/Microsoft.AspNetCore.Rewrite/Extensions/HttpRequestExtensions.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Rewrite.Extensions
+{
+    public static class HttpRequestExtensions
+    {
+        public static Uri ToUri(this HttpRequest request)
```
Tratcher: This is only used in tests, move it to the test project.

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/IISUrlRewriteRule.cs
```diff
@@ -57,6 +65,22 @@ public virtual void ApplyRule(RewriteContext context)
                 }
             }
 
+            foreach (ServerVariable serverVariable in ServerVariables)
+            {
+                if (serverVariable.Name.StartsWith(RequestHeaderPrefix, StringComparison.CurrentCultureIgnoreCase))
```
Tratcher: Ordinal  ignore case

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/IISUrlRewriteRule.cs
```diff
@@ -57,6 +65,22 @@ public virtual void ApplyRule(RewriteContext context)
...
+                {
+                    context.HttpContext.Request.Headers.Add(
+                        serverVariable.Name.Substring(RequestHeaderPrefix.Length).Replace('_', '-'),
+                        serverVariable.Evaluate(context, initMatchResults, condMatchRes));
+                }
+                else if (serverVariable.Name.StartsWith(ResponseHeaderPrefix, StringComparison.CurrentCultureIgnoreCase))
```
Tratcher: Ordinal ignore case

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/IISUrlRewriteRule.cs
```diff
@@ -57,6 +65,22 @@ public virtual void ApplyRule(RewriteContext context)
                 }
             }
 
+            foreach (ServerVariable serverVariable in ServerVariables)
+            {
+                if (serverVariable.Name.StartsWith(RequestHeaderPrefix, StringComparison.CurrentCultureIgnoreCase))
+                {
+                    context.HttpContext.Request.Headers.Add(
```
Tratcher: Append

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/IISUrlRewriteRule.cs
```diff
@@ -57,6 +65,22 @@ public virtual void ApplyRule(RewriteContext context)
...
+                        serverVariable.Name.Substring(RequestHeaderPrefix.Length).Replace('_', '-'),
+                        serverVariable.Evaluate(context, initMatchResults, condMatchRes));
+                }
+                else if (serverVariable.Name.StartsWith(ResponseHeaderPrefix, StringComparison.CurrentCultureIgnoreCase))
+                {
+                    context.HttpContext.Response.Headers.Add(
```
Tratcher: Append

davidpeden3:
@jkotalik i have them all implemented in an actual reverse proxy that we're running and implemented everything that we needed to migrate away from IIS/ARR. as you have noted elsewhere, documentation on how things should work is sparse so i've emulated the behavior we've observed in IIS/ARR and written tests to back that up. undoubtedly there will be holes but the PRs should at least get the ball rolling.

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
```diff
@@ -142,6 +143,33 @@ private void ParseCondition(XElement condition, UrlRewriteRuleBuilder builder, P
...
+            {
+                ParseServerVariable(serverVariable, builder);
+            }
+        }
+
+        private void ParseServerVariable(XElement serverVariable, UrlRewriteRuleBuilder builder)
+        {
+            try
+            {
+                builder.AddServerVariable(
```
Tratcher: Throw NotSupported here if Name is not one of the supported variables HTTP_ or RESPONSE_.

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
```diff
@@ -142,6 +143,33 @@ private void ParseCondition(XElement condition, UrlRewriteRuleBuilder builder, P
...
+                    serverVariable.Attribute(RewriteTags.Name).Value,
+                    _inputParser.ParseInputString(serverVariable.Attribute(RewriteTags.Value).Value));
```
jkotalik: Either note or add the ability to support the Replace flag, which will replace the server variable if it already exists if true, else will throw or not replace it. To support it, simply add a check to UrlRewriteRuleBuilder to check if the list already contains a value name.

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteRuleBuilder.cs
```diff
@@ -19,6 +19,7 @@ public class UrlRewriteRuleBuilder
 
         private UrlMatch _initialMatch;
         private IList<Condition> _conditions;
+        private readonly IList<ServerVariable> _serverVariables = new List<ServerVariable>();
```
jkotalik: Based on previous comment with supporting the Replace tag, I'd make this a dict from name to pattern.

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/IISUrlRewriteRule.cs
```diff
@@ -57,6 +65,22 @@ public virtual void ApplyRule(RewriteContext context)
                 }
             }
 
+            foreach (ServerVariable serverVariable in ServerVariables)
+            {
+                if (serverVariable.Name.StartsWith(RequestHeaderPrefix, StringComparison.CurrentCultureIgnoreCase))
+                {
+                    context.HttpContext.Request.Headers.Add(
+                        serverVariable.Name.Substring(RequestHeaderPrefix.Length).Replace('_', '-'),
```
jkotalik: The replace should be done in parsing because we will know the name.

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/IISUrlRewriteRule.cs
jkotalik: Rather than doing string comparisons on the name, we can probably have an enum with different types (RequestHeaderPrefix, ResponseHeaderPrefix, etc.)

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/IISUrlRewriteRule.cs
```diff
@@ -57,6 +65,22 @@ public virtual void ApplyRule(RewriteContext context)
...
+                        serverVariable.Evaluate(context, initMatchResults, condMatchRes));
+                }
+                else if (serverVariable.Name.StartsWith(ResponseHeaderPrefix, StringComparison.CurrentCultureIgnoreCase))
+                {
+                    context.HttpContext.Response.Headers.Add(
+                        serverVariable.Name.Substring(ResponseHeaderPrefix.Length).Replace('_', '-'),
```
jkotalik: Same as above.

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/IISUrlRewriteRule.cs
```diff
@@ -57,6 +65,22 @@ public virtual void ApplyRule(RewriteContext context)
...
+                        serverVariable.Evaluate(context, initMatchResults, condMatchRes));
```
jkotalik: Would be great if we can get some logging in here when we we add a server variable (less of a priority).

test/Microsoft.AspNetCore.Rewrite.Tests/IISUrlRewrite/FileParserTests.cs
```diff
@@ -147,7 +148,7 @@ public void RuleParse_ParseMultipleRules()
             )
         {
             return new IISUrlRewriteRule(name, new RegexMatch(new Regex("^OFF$"), false), conditions,
-                new RewriteAction(RuleResult.ContinueRules, new InputParser().ParseInputString(url), queryStringAppend: false));
+                Enumerable.Empty<ServerVariable>(), new RewriteAction(RuleResult.ContinueRules, new InputParser().ParseInputString(url), queryStringAppend: false));
```
jkotalik: @Tratcher We should definitely have more tests here.

test/Microsoft.AspNetCore.Rewrite.Tests/IISUrlRewrite/MiddleWareTests.cs
```diff
@@ -366,5 +371,279 @@ public class MiddlewareTests
 
             Assert.Equal("/hey/hello/", response.Headers.Location.OriginalString); ;
         }
+
```
jkotalik: Good tests, I'll look into this once we get the functional code wrapped up.

test/Microsoft.AspNetCore.Rewrite.Tests/project.json
```diff
@@ -5,7 +5,10 @@
   },
   "dependencies": {
     "dotnet-test-xunit": "2.2.0-*",
-    "Microsoft.AspNetCore.Rewrite": "1.0.0-*",
+			"Microsoft.AspNetCore.Rewrite": {
```
jkotalik: I believe this is causing the build error?

Tratcher: Yeah, revert this

Tratcher: You may need to rebase: https://github.com/aspnet/BasicMiddleware/commit/ea0042c685ba2228d0d4b85df41fa45a0b360f00

src/Microsoft.AspNetCore.Rewrite/Extensions/HttpRequestExtensions.cs
davidpeden3: @Tratcher it's also used in my [global rules PR](https://github.com/aspnet/BasicMiddleware/pull/169/commits/d99270576e9fb996ffec40d52ed74a8d3f623cb6)

davidpeden3:
@Tratcher @jkotalik i made the easy changes. i probably won't be able to get to the more involved stuff until wed or next week, unfortunately.

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/IISUrlRewriteRule.cs
davidpeden3: i did migrate them into a constants file but not following your suggestion here. can you clarify?

davidpeden3:
ok, was able to knock out everything else. let me know if i missed anything.

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/IISUrlRewriteRule.cs
```diff
@@ -1,7 +1,9 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
+
```
mikaelm12: nit: Remove newline

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/ServerVariable.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
```
mikaelm12: Copyright and license header here

src/Microsoft.AspNetCore.Rewrite/Extensions/HttpRequestExtensions.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
```
jkotalik: Copyright header

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/IISUrlRewriteRule.cs
```diff
@@ -57,6 +62,29 @@ public virtual void ApplyRule(RewriteContext context)
                 }
             }
 
+            foreach (ServerVariable serverVariable in ServerVariables)
```
jkotalik: Probably should check ServerVariables is not null or Empty

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
```diff
@@ -142,6 +143,53 @@ private void ParseCondition(XElement condition, UrlRewriteRuleBuilder builder, P
...
+                string prefix;
+                ServerVariableType type;
+                var name = serverVariable.Attribute(RewriteTags.Name).Value;
+
+                if (name.StartsWith(ServerVariableConstants.RequestHeaderPrefix, StringComparison.OrdinalIgnoreCase))
+                {
+                    prefix = ServerVariableConstants.RequestHeaderPrefix;
+                    type = ServerVariableType.Request;
+                }
+                else if (name.StartsWith(ServerVariableConstants.ResponseHeaderPrefix, StringComparison.OrdinalIgnoreCase))
```
jkotalik: @Tratcher @mikaelm12 @davidpeden3 See https://www.iis.net/learn/extensions/url-rewrite-module/url-rewrite-module-20-configuration-reference#Setting_Server_Variables with response headers. I think they must be with outbound rules, so until we implement outbound rules, lets remove response headers?

davidpeden3: IIS/ARR definitely allow setting a response header as a part of an inbound rule. i emulated the behavior with the configuration we have in `applicationHost.config`. i don't have any outbound rules in my config otherwise i would've implemented that feature as well. ;)

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/ServerVariableConstants.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Microsoft.AspNetCore.Rewrite.Internal.IISUrlRewrite
```
davidpeden3: style-wise, do you prefer a separate file or would these be better as public constants directly on `ServerVariable`?

davidpeden3: i made the change as requested. let me know if you want the constants moved.

src/Microsoft.AspNetCore.Rewrite/Extensions/HttpRequestExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Extensions;
+
+namespace Microsoft.AspNetCore.Rewrite.Extensions
+{
+    public static class HttpRequestExtensions
+    {
+        public static string ToAbsoluteUri(this HttpRequest request)
```
Tratcher: Isn't this a duplicate of GetEncodedUrl? https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http.Extensions/UriHelper.cs#L166.
Also, it's only used in tests.

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
Tratcher: The docs say it's only for outbound rules. @mikaelm12 can you do some manual tests against the IIS module and see if this lines up?

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/IISUrlRewriteRule.cs
```diff
@@ -57,6 +61,32 @@ public virtual void ApplyRule(RewriteContext context)
...
+                            break;
+                        case ServerVariableType.Response:
+                            headerDictionary = context.HttpContext.Response.Headers;
+                            context.Logger?.ResponseHeaderAdded(name, value);
+                            break;
+                        default:
+                            throw new ArgumentOutOfRangeException($"Unexpected server variable: `{serverVariable.Type}`");
+                    }
+
+                    headerDictionary.Append(name, value);
```
Tratcher: @mikaelm12 this is another behavior we need to verify against the naitive module, What happens when the header already exists? Replace? Append?

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/ServerVariableType.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Rewrite.Internal.IISUrlRewrite
+{
+    public enum ServerVariableType
+    {
+        Request,
```
Tratcher: RequestHeader & ResponseHeader?

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/ServerVariableConstants.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Rewrite.Internal.IISUrlRewrite
+{
+    public static class ServerVariableConstants
+    {
+        public const string RequestHeaderPrefix = "HTTP_";
+        public const string ResponseHeaderPrefix = "RESPONSE_";
```
Tratcher: There are a bunch more here: https://github.com/aspnet/BasicMiddleware/blob/623333a919c34e63ae98a276564e03f31f9cb1cf/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/ServerVariables.cs#L17

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
```diff
@@ -142,6 +143,53 @@ private void ParseCondition(XElement condition, UrlRewriteRuleBuilder builder, P
...
+
+        private void ParseServerVariable(XElement serverVariable, UrlRewriteRuleBuilder builder)
+        {
+            try
+            {
+                string prefix;
+                ServerVariableType type;
+                var name = serverVariable.Attribute(RewriteTags.Name).Value;
+
+                if (name.StartsWith(ServerVariableConstants.RequestHeaderPrefix, StringComparison.OrdinalIgnoreCase))
```
Tratcher: Our server variable lookup isn't this generic but probably should be.
https://github.com/aspnet/BasicMiddleware/blob/623333a919c34e63ae98a276564e03f31f9cb1cf/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/ServerVariables.cs#L25

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteRuleBuilder.cs
```diff
@@ -149,5 +150,10 @@ public void AddUrlConditions(LogicalGrouping logicalGrouping, bool trackAllCaptu
             _conditions = new List<Condition>();
             _matchAny = logicalGrouping == LogicalGrouping.MatchAny;
         }
+
+        public void AddOrUpdateServerVariable(string name, Pattern pattern, ServerVariableType type)
```
Tratcher: SetServerVariable?

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/IISUrlRewriteRule.cs
davidpeden3: two of my tests started failing after this change. i decompiled append and see that it no-ops if the key doesn't already exist. going to change the code to add or append as necessary.

src/Microsoft.AspNetCore.Rewrite/Extensions/HttpRequestExtensions.cs
davidpeden3: yep. just plain missed it. fixed

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/IISUrlRewriteRule.cs
```diff
@@ -83,7 +83,14 @@ public virtual void ApplyRule(RewriteContext context)
                             throw new ArgumentOutOfRangeException($"Unexpected server variable: `{serverVariable.Type}`");
                     }
 
-                    headerDictionary.Append(name, value);
+                    if (headerDictionary.ContainsKey(name))
```
Tratcher: ???
https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http.Abstractions/Internal/ParsingHelpers.cs#L158

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
davidpeden3: @Tratcher are you suggesting i make a change or just merely making an observation?

Tratcher: Try to make the change yourself, or file a bug for later followup.

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/IISUrlRewriteRule.cs
davidpeden3: i had to pull in the source to see what was happening. `.Append` is fine as you pointed out. the reason `.Append` fails my tests is because of what i noted in the test:

```csharp
// REMOTE_ADDR is not set by the test framework, so we assert the existence of the header rather than its value
```

which causes this to run:

https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http.Abstractions/Internal/ParsingHelpers.cs#L104

so i'll change the tests to pass and restore `.Append`.

is there a way to get `TestServer` to actually hydrate `HttpContext.Connection.RemoteIpAddress`?

Tratcher: No, TestServer doesn't support RemoteIpAddress. The best you can do is fill it in via an inline middleware:
```
app.Use((context, next) =>
{
  context.Connection.RemoteIpAddress = ...;
  return next();
});

File: src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs (hunk 26 above)
Comment (davidpeden3) [inline src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs]: ok, not sure if https://github.com/aspnet/BasicMiddleware/pull/184/commits/ded8bbc036133a0a29a48928c780b7597016d3ce is what you had in mind. let me know what you think.

File: src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/ServerVariables.cs [hunk 29]
```diff
@@ -19,44 +26,75 @@ public static PatternSegment FindServerVariable(string serverVariable, ParserCon
...
+                    break;
                 case "REQUEST_URI":
-                    return new UrlSegment();
+                    patternSegment = new UrlSegment();
+                    break;
                 default:
                     throw new FormatException(Resources.FormatError_InputParserUnrecognizedParameter(serverVariable, context.Index));
             }
+
+            return new ServerVariable(serverVariable, new Pattern(patternSegment), ServerVariableType.RequestHeader);
```
Tratcher: ? Only some of these variables are request headers.

davidpeden3:
@Tratcher back from vacation and was going to resume work on these PRs. started off with attempting to sync my fork with `dev`. i've run into a snag.

https://github.com/aspnet/BasicMiddleware/blob/dev/global.json#L6

does not appear to be a publicly available version of the sdk. i am running `1.0.0-preview2-1-003177` locally and that's what's available from:

https://www.microsoft.com/net/download/core#/current

i tried just changing the value to build `3177` but i get over 200 compilation errors related to `Task`. E.g.,:

```Severity	Code	Description	Project	File	Line	Suppression State
Error	CS0433	The type 'Task' exists in both 'System.Threading.Tasks, Version=4.0.10.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' and 'System.Runtime, Version=4.2.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'	Microsoft.AspNetCore.HttpOverrides.Tests	C:\temp\BasicMiddleware\test\Microsoft.AspNetCore.HttpOverrides.Tests\ForwardedHeadersMiddlewareTest.cs	18	
```

None of which are in my code. This also happens on a fresh clone of `dev` that does not have any of my code. I'm not sure what the issue is nor how to proceed. Any advice would be appreciated.

Tratcher:
Working against the dev can be hard when we take the latest CLI builds. Start by running build.cmd from the command line. That is self-contained and pulls all of the dependencies it needs.  If that works then you can modify your machine path to remove `C:\Program Files\dotnet` and add `C:\Users\{username}\appdata\local\Microsoft\dotnet`. Then VS should be able to find the latest CLI version installed by build.cmd.

davidpeden3:
great, thanks. i'll give that a shot.

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/ServerVariables.cs
davidpeden3: need some feedback on which are not and what they are instead and then i can dial this in.

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/IISUrlRewriteRule.cs
mikaelm12: There's an option in the IIS Url Rewrite module which is set by default to replace the existing request header value if it already exists. Otherwise it seems as if it leaves it alone.

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/ServerVariables.cs
davidpeden3: @Tratcher is this the only outstanding item for this PR? i'll make the necessary change(s) if you can provide some guidance. thanks.

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/IISUrlRewriteRule.cs
Tratcher: Still need to address this one too

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/ServerVariables.cs
Tratcher: Only the HTTP_* entries are request headers, the ones that use HeaderSegment. All of the others are for different fields on the request. Your logic in IISUrlRewriteRule doesn't account for all of the other types.

davidpeden3: Yeah, understood. What I'm not clear on as how the non-request header server variables should be classified. The closest thing to a classification that I see is from [the docs](https://msdn.microsoft.com/en-us/library/ms524602(v=vs.90).aspx):

```
IIS server variables provide information about the server, the connection with the client, and the current request on the connection.
```

I'm not really sure how that should translate to the code. Request/response header types were obvious and necessary. What do you recommend for the other server variables?

Is just adding `ServerVariableType.Request` sufficient? Does that capture the domain adequately?

davidpeden3: https://github.com/aspnet/BasicMiddleware/pull/184/commits/80a454cb724178453b1eb1371b75f6e3eb39669b

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/IISUrlRewriteRule.cs
davidpeden3: https://github.com/aspnet/BasicMiddleware/pull/184/commits/696c63b8a818782a24ab584b6fc61a169bc11c2a

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/ServerVariables.cs
```diff
@@ -9,16 +9,21 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.IISUrlRewrite
...
+        /// Returns a <see cref="ServerVariable"/>
         /// </summary>
         /// <param name="serverVariable">The server variable</param>
         /// <param name="context">The parser context which is utilized when an exception is thrown</param>
         /// <param name="global">Indicates if the rule being parsed is a global rule</param>
         /// <exception cref="FormatException">Thrown when the server variable is unknown</exception>
-        /// <returns>The matching <see cref="PatternSegment"/></returns>
-        public static PatternSegment FindServerVariable(string serverVariable, ParserContext context, bool global)
+        /// <returns>A <see cref="ServerVariable"/></returns>
+        public static ServerVariable FindServerVariable(string serverVariable, ParserContext context, bool global)
```
Tratcher: I think the problem here is that we're conflating two independent scenarios. Server Variables can be read from or assigned to. Previously this code only supported reading. Now you're trying to add assignment, but you've only implemented two of the variable types, request and response headers.

I think this needs to be split up. Revert the changes to FindServerVariable and create a new method FindServerVariableSetter. FindServerVariableSetter looks up the variables and returns a new kind set of types that mutate the request/response when evaluated. E.g. you need to create a type that's the opposite of HeaderSegment, a HeaderSetter, and a UrlSetter, SchemeSetter, etc.. Then when you loop through these Setters in ApplyRule, you don't need a switch statement, you just call Apply on each one.

davidpeden3:
@mikaelm12 i have not resolved the last comment that @Tratcher made. been tied up at work and need to chew on it some more to better understand what changes i should make.

Issue #196, opened 2016-12-25
Design Question on URL Rewriting Middleware


guardrex:
`.AddRedirect()` and `.AddRewrite()` match their regex on the path/resource (not including the scheme, host, port, and query).

Works ...
```
.AddRedirect(@"^rule1/(.*)", "https://www.google.com/?query=$1")
.AddRedirect(@"^rule2/(.*)", "rule2/?query=$1")
```
Doesn't work ...
```
.AddRedirect(@"^http://localhost:5002/rule3/(.*)", "https://www.google.com/?query=$1")
```
Might be because "this is always the way it has always been done," but why is the design such that matching isn't performed on the absolute Url? ... or that there isn't an option to instruct the middleware to match on the absolute Url ... for maximum flexibility.

davidpeden3:
@GuardRex i didn't (think to) add any support to the methods you mention above in https://github.com/aspnet/BasicMiddleware/pull/169 but can/should? that's precisely the kind of support i needed/added in the file-based configuration.

davidpeden3:
Looks like this is the same issue as https://github.com/aspnet/BasicMiddleware/issues/193 and https://github.com/aspnet/BasicMiddleware/issues/194.

jkotalik:
@muratg is this a doc issue at this point?

muratg:
@jkotalik I believe so. @shirhatti is that the case?

Issue #174, opened 2016-11-09
Support simple redirect rules


strohhut:
What if I want case insensitive matching? What if I want to match https requests so I can redirect them to http (or vice-versa)? What if I want to redirect www to non-www (or vice-versa)?

Do I really have to implement my own IRule for these simple use cases? The `RewriteRule` class seems much too basic for me.

strohhut:
I found that with `(?i)` case insesitive matching without `RegexOptions.IgnoreCase` can be done. 

Please also consider port and query dependent redirects besides domain and protocol ones I described above.

Tratcher:
@strohhut thanks for the input. We're still working on getting the core infrastructure finished and then we can explore more of these scenarios.

You can also do some of these things using the UrlRewrite and ModRewrite infrastructure already added.

They way the infrastructure is setup it should be pretty easy for people to build and share custom code rules. https://github.com/aspnet-contrib may be a good place for collaberating on these.

strohhut:
Thx, for the help and considering this. `AddIISUrlRewrite` surely is a good idea if you need more than `RedirectRule` provides and if you don't want to implement your own rule classes.

muratg:
This needs design. Putting in 1.2.0 for now.

PR #198, opened 2017-01-10
Make condition collection a first class citizen

davidpeden3:
This PR builds on top of https://github.com/aspnet/BasicMiddleware/pull/169 and cleans up the conditions model. It adds a dedicated `ConditionsCollection`.


src/Microsoft.AspNetCore.Rewrite/project.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "1.1.0-*",
+  "version": "1.2.0-*",
```
natemcmaster: At the moment, this versioning is deliberate. Can you leave this as-is?

davidpeden3: done. also applied in my other PRs.

davidpeden3:
@Tratcher i've branched this work off of the global rules branch. if you like this, i'd prefer to roll it in to my global rules branch asap so that i'm not maintaining a branch of a branch.

src/Microsoft.AspNetCore.Rewrite/IISUrlRewriteOptionsExtensions.cs
```diff
@@ -65,5 +65,17 @@ public static RewriteOptions AddIISUrlRewrite(this RewriteOptions options, TextR
...
             return options;
         }
+
+        /// <summary>
+        /// Adds a rule to the current rules.
+        /// </summary>
+        /// <param name="options">The <see cref="RewriteOptions"/>.</param>
+        /// <param name="rule">A <see cref="IISUrlRewriteRule"/> to be added to the current rules.</param>
+        /// <returns>The Rewrite options.</returns>
+        public static RewriteOptions AddIISUrlRewrite(this RewriteOptions options, IISUrlRewriteRule rule)
```
Tratcher: This extension is redundant with Add(IRule)

samples/RewriteSample/Startup.cs
```diff
@@ -19,13 +20,61 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
...
+        ///     <match url="(.*)" ignoreCase="false" />
+        ///     <conditions trackAllCaptures = "true" >
+        ///         <add input="{REQUEST_URI}" pattern="^/([a-zA-Z]+)/([0-9]+)$" />
+        ///         <add input="{QUERY_STRING}" pattern="p2=([a-z]+)" />
+        ///     </conditions>
+        ///     <action type="Redirect" url ="blogposts/{C:1}/{C:4}" />
+        /// </rule>
+        /// </summary>
+        /// <returns>An <see cref="IISUrlRewriteRule"/></returns>
+        private static IISUrlRewriteRule CreateCustomIISRule()
```
Tratcher: :-1: IISUrlRewriteRule isn't intended for developers to construct directly, it's just the object model for representing the rules in the xml file. Yes a dev can still do this, but it's not something we want to show in the samples.

src/Microsoft.AspNetCore.Rewrite/IISUrlRewriteOptionsExtensions.cs
davidpeden3: Ah, good point. I'll take it out.

samples/RewriteSample/Startup.cs
davidpeden3: That's unfortunate. I specifically need to build some rules from an external configuration (e.g., from a service or stored in a database). I would imagine this isn't that uncommon of a scenario. But, ok.

Tratcher: Feel free to use the infrastructure, but it's not something we'll recommend in the samples.

davidpeden3:
@Tratcher 

Sure, sorry. I had been heads-down in the code for a bit did not explain the context I had in my head. Of course this PR needs more clarity.

This branch off of the `globalrules` branch does two things:

First, it addresses the `Condition` abstraction issues I've gotten to know while working in the codebase.

`Condition`s are passed around as an `IList` and expose an `OrNext` property. The only reason this property exists on `Condition` is to support `ConditionHelper.Evaluate`. In the absence of a proper collection which should contain this property (as well as `TrackAllCaptures`), the condition itself maintains state. This PR creates a dedicated collection for `Condition`s and also houses the grouping and `trackAllCaptures` at the appropriate level which cleans up a lot of the related classes.

e.g.,:

Each individual `Condition` should not expose the behavior of the collection. With this implementation, it's possible to define different values per `Condition` which is invalid via file-based configuration:

https://github.com/aspnet/BasicMiddleware/pull/198/commits/347d7bd455e9b536422f48006553e2714f894f02#diff-b9bceaec9393b4ce7c2e1a9057e1ad9fL22

It doesn't make any sense to have to pass `trackAllCaptures` to a method that parses a `Condition`:

https://github.com/aspnet/BasicMiddleware/pull/198/files#diff-4e6cd3225a4ad73df916578745fa9724L121

Second, when I originally set out to implement global rules, I was literally translating the observed behavior we had configured in IIS to this middleware. For global rules, that turned out to be that the full URI was evaluated for matches, not just the path. I built that without much consideration to the rest of the codebase (I simply wasn't as familiar with it as I am today).

But it turns out that there is demand for greater flexibility than that and that it would be great for local (non-global) rules to also be able to benefit from full URI comparison:

https://github.com/aspnet/BasicMiddleware/issues/174

and

https://github.com/aspnet/BasicMiddleware/issues/196

So, it made more sense to me to adjust the code to reflect matching URIs on a path/full basis explicitly rather than implicitly from global rules.

src/Microsoft.AspNetCore.Rewrite/IISUrlRewriteOptionsExtensions.cs
davidpeden3: I was about to fix these two items but am not sure how to get runtime 1.2.0-beta-001289-00 now required by this commit: https://github.com/aspnet/KoreBuild/commit/27d74d6e4ef457ce8d195c2506161582064bbf41.

Running `build.cmd` doesn't seem to fetch it like it did originally for me. Here's what I see now:

```The specified framework 'Microsoft.NETCore.App', version '1.2.0-beta-001289-00' was not found.
  - Check application dependencies and target a framework version installed at:
      C:\Users\dpeden\AppData\Local\Microsoft\dotnet\shared\Microsoft.NETCore.App\
  - The following versions are installed:
      1.0.0
      1.1.0
      1.2.0-beta-001269-00
      1.2.0-beta-001285-00
  - Alternatively, install the framework version '1.2.0-beta-001289-00'.
```

I'll go ahead and push but I can't run tests without this framework. Any ideas on how to fix this?

Tratcher: build.cmd is lazy, sometimes you need to delete the cached files before it will download new ones. `git clean -xdf`

davidpeden3: bingo

src/Microsoft.AspNetCore.Rewrite/Internal/ConditionCollection.cs
```diff
@@ -0,0 +1,84 @@
...
+            {
+                if (index < _conditions.Count)
+                {
+                    return _conditions[index];
+                }
+                throw new IndexOutOfRangeException($"Cannot access condition at index {index}. Only {_conditions.Count} conditions were captured.");
+            }
+        }
+
+        public void Add(Condition condition)
```
mikaelm12: An `AddConditions` method might be nice to add here

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteRuleBuilder.cs
```diff
@@ -86,70 +84,40 @@ public void AddUrlMatch(string input, bool ignoreCase = true, bool negate = fals
...
+        public void AddUrlConditions(IEnumerable<Condition> conditions)
         {
-            _conditions = new List<Condition>();
-            _matchAny = logicalGrouping == LogicalGrouping.MatchAny;
-            _trackAllCaptures = trackAllCaptures;
+            if (_conditions == null)
+            {
+                throw new InvalidOperationException($"You must first configure condition behavior by calling {nameof(ConfigureConditionBehavior)}");
+            }
+            foreach (var condition in conditions)
```
mikaelm12: Could replace this with `AddConditions` mentioned above

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
```diff
@@ -5,9 +5,12 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Xml;
 using System.Xml.Linq;
```
mikaelm12: Remove new line

src/Microsoft.AspNetCore.Rewrite/Internal/ConditionCollection.cs
davidpeden3: https://github.com/aspnet/BasicMiddleware/pull/198/commits/3f241c62c8ec496b8fe37ae999b250a3af6d7b9e

src/Microsoft.AspNetCore.Rewrite/Internal/ConditionCollection.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Rewrite.Internal
+{
+    public class ConditionCollection : IEnumerable<Condition>
+    {
+        private readonly List<Condition> _conditions = new List<Condition>();
+
+        public enum ConditionGrouping
```
Tratcher: just use LogicalGrouping

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteRuleBuilder.cs
```diff
@@ -18,19 +18,17 @@ public class UrlRewriteRuleBuilder
...
 
         private UrlMatch _initialMatch;
-        private IList<Condition> _conditions;
+        private ConditionCollection _conditions;
         private UrlAction _action;
-        private bool _matchAny;
-        private bool _trackAllCaptures;
 
-        public IISUrlRewriteRule Build()
+        public IISUrlRewriteRule Build(bool global)
```
Tratcher: Why not make global a property on the rule builder?

src/Microsoft.AspNetCore.Rewrite/Internal/ApacheModRewrite/ApacheModRewriteRule.cs
```diff
@@ -9,10 +9,10 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.ApacheModRewrite
     public class ApacheModRewriteRule : IRule
     {
         public UrlMatch InitialMatch { get; }
-        public IList<Condition> Conditions { get; }
+        public ConditionCollection Conditions { get; }
```
Tratcher: mod rewrite uses different conditional logic where the and/or is per condition. What you're doing to UrlRewrite does not apply. Revert.

src/Microsoft.AspNetCore.Rewrite/Internal/ConditionCollection.cs
```diff
@@ -0,0 +1,92 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Rewrite.Internal
```
Tratcher: move to IISUrlRewrite namespace and directory, this is not a shared component.

src/Microsoft.AspNetCore.Rewrite/Internal/ConditionCollection.cs
```diff
@@ -0,0 +1,92 @@
...
+        {
+        }
+
+        public ConditionCollection(ConditionGrouping grouping, bool trackAllCaptures)
+        {
+            Grouping = grouping;
+            TrackAllCaptures = trackAllCaptures;
+        }
+
+        public ConditionCollection(ConditionGrouping grouping, bool trackAllCaptures, Condition condition)
```
Tratcher: I don't think the 3rd and 4th constructor overloads are being used anywhere. Remove.

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/IISUrlRewriteRule.cs
```diff
@@ -11,21 +10,21 @@ public class IISUrlRewriteRule : IRule
...
-        public bool TrackAllCaptures { get; }
+        public bool Global { get; }
 
         public IISUrlRewriteRule(string name,
             UrlMatch initialMatch,
-            IList<Condition> conditions,
+            ConditionCollection conditions,
             UrlAction action,
-            bool trackAllCaptures)
+            bool global = false)
```
Tratcher: avoid default parameters. add a second constructor if you need it.

PR #199, opened 2017-01-11
Add Vary: Accept-Encoding if response was compressed

### LINKED ISSUE #187: Compression: Vary Header not returned when content gzipped
Response header should contain
```
Vary: Accept-Encoding
```

when content is gzipped. IIS compression currently adds header.

JunTaoLuo:
Fixes #187

### DIFF
--- a/src/Microsoft.AspNetCore.ResponseCompression/BodyWrapperStream.cs
+++ b/src/Microsoft.AspNetCore.ResponseCompression/BodyWrapperStream.cs
@@ -211,6 +211,24 @@ namespace Microsoft.AspNetCore.ResponseCompression
                 _compressionChecked = true;
                 if (_provider.ShouldCompressResponse(_context))
                 {
+                    // If the MIME type indicates that the response could be compressed, caches will need to vary by the Accept-Encoding header
+                    var varyValues = _context.Response.Headers.GetCommaSeparatedValues(HeaderNames.Vary);
+                    var varyByAcceptEncoding = false;
+
+                    for (var i = 0; i < varyValues.Length; i++)
+                    {
+                        if (string.Equals(varyValues[i], HeaderNames.AcceptEncoding, StringComparison.OrdinalIgnoreCase))
+                        {
+                            varyByAcceptEncoding = true;
+                            break;
+                        }
+                    }
+
+                    if (!varyByAcceptEncoding)
+                    {
+                        _context.Response.Headers.Append(HeaderNames.Vary, HeaderNames.AcceptEncoding);
+                    }
+
                     var compressionProvider = ResolveCompressionProvider();
                     if (compressionProvider != null)
                     {
--- a/test/Microsoft.AspNetCore.ResponseCompression.Tests/BodyWrapperStreamTests.cs
+++ b/test/Microsoft.AspNetCore.ResponseCompression.Tests/BodyWrapperStreamTests.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
+using Microsoft.Net.Http.Headers;
 using Moq;
 using Xunit;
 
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNetCore.ResponseCompression/BodyWrapperStream.cs
```diff
@@ -214,6 +214,11 @@ private void OnWrite()
                     var compressionProvider = ResolveCompressionProvider();
                     if (compressionProvider != null)
                     {
+                        if (!HeaderUtilities.ContainsCacheDirective(HeaderNames.Vary, HeaderNames.AcceptEncoding))
+                        {
+                            _context.Response.Headers.Add(HeaderNames.Vary, HeaderNames.AcceptEncoding);
```
Tratcher: Append or AppendCommaSeperatedValue

src/Microsoft.AspNetCore.ResponseCompression/BodyWrapperStream.cs
```diff
@@ -214,6 +214,11 @@ private void OnWrite()
                     var compressionProvider = ResolveCompressionProvider();
                     if (compressionProvider != null)
                     {
+                        if (!HeaderUtilities.ContainsCacheDirective(HeaderNames.Vary, HeaderNames.AcceptEncoding))
```
JunTaoLuo: This utility is named poorly but does what is needed which is to scan the values of the headers and see if the specific value, in this case Accept-Encoding, exist.

Tratcher: File a bug for it

JunTaoLuo: Actually what I said previously isn't accurate, the logic in HeaderUtilities is specific for cache directives. This is because it only handles the cases where the header values are in the format `valueName=<someValue>, ...` which doesn't generalize to other headers such as Accept where the headers can look like `valueName;q=<qValue>, ...`. For the use case here, the layout of the header value is even simpler since it is a comma separated list of tokens. I'll add a separate helper function.

test/Microsoft.AspNetCore.ResponseCompression.Tests/ResponseCompressionMiddlewareTest.cs
```diff
@@ -793,13 +793,24 @@ private void CheckResponseCompressed(HttpResponseMessage response, int expectedB
         {
             IEnumerable<string> contentMD5 = null;
 
+            var containsVaryAcceptEncoding = false;
+            foreach (var value in response.Headers.GetValues(HeaderNames.Vary))
+            {
+                if (value.Contains(HeaderNames.AcceptEncoding))
```
Tratcher: why a Contains vs. an exact match?

Tratcher: Or are you trying to avoid parsing the header into multiple segments?

JunTaoLuo: In this case either would work since the tests only set one value to the Vary header but in general when we are dealing with headers, we can't just assume each string in StringValues contains one value. We ran into this problem when we received Header["Cache-Control"] = new[] {"no-cache,no-store"} from MVC for example. So it's best to check using contains rather than exact match. 

In the same regard, we also shouldn't be assuming the format of the headers. For example, the test could break with a bad Vary header like Headers["Vary"] = new [] {"badValue=Accept-Encoding"} but I ignored this corner case in the tests.

test/Microsoft.AspNetCore.ResponseCompression.Tests/BodyWrapperStreamTests.cs
```diff
@@ -9,12 +9,32 @@
 using Microsoft.AspNetCore.Http;
 using Moq;
 using Xunit;
+using Microsoft.Net.Http.Headers;
```
Tratcher: Sort

JunTaoLuo:
Tested with the response caching middleware and filed a bug https://github.com/aspnet/ResponseCaching/issues/92. Currently if the response is compressed before being cached, the expected behaviour where the response is cached per accept encoding is observed. If the compression comes after caching, however, the response is still cached per accept encoding. This issue will be fixed in the response caching middleware.

PR #206, opened 2017-02-10
async all the things

davidpeden3:


src/Microsoft.AspNetCore.Rewrite/Internal/UrlActions/CustomResponseAction.cs
```diff
@@ -34,7 +35,7 @@ public override void ApplyAction(RewriteContext context, BackReferenceCollection
                 var content = Encoding.UTF8.GetBytes(StatusDescription);
                 response.ContentLength = content.Length;
                 response.ContentType = "text/plain; charset=utf-8";
-                response.Body.Write(content, 0, content.Length);
+                await response.Body.WriteAsync(content, 0, content.Length);
```
Tratcher: async/await can be avoided if you re-order

src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/IISUrlRewriteRule.cs
```diff
@@ -74,7 +75,7 @@ public virtual void ApplyRule(RewriteContext context)
 
             context.Logger?.UrlRewriteMatchedRule(Name);
             // at this point we know the rule passed, evaluate the replacement.
-            Action.ApplyAction(context, initMatchResults?.BackReferences, condResult?.BackReferences);
+            await Action.ApplyActionAsync(context, initMatchResults?.BackReferences, condResult?.BackReferences);
```
davidfowl: Remove the async/await here and just return the task.

davidfowl:
> @davidfowl let's not, this is going to make the perf even worse. 

Let's see some measurements before making those claims. Caching non generic tasks are super easy. As soon as we get rules that read the request body or write to the response body then we need an async option.

> Also, it's a pretty big breaking change for anyone that's already implemented rules.

We *can* make breaking changes in 2.0 and my guess is that we wrote most of the rules since this middleware hasn't been around that wrong.

davidpeden3:
@Tratcher is this what you were suggesting?

https://gist.github.com/davidpeden3/1ef73ea305fe262da5aab60150219131

src/Microsoft.AspNetCore.Rewrite/Internal/ApacheModRewrite/ApacheModRewriteRule.cs
```diff
@@ -20,36 +22,33 @@ public ApacheModRewriteRule(UrlMatch initialMatch, IList<Condition> conditions,
...
 
             // At this point, we know our rule passed, first apply pre conditions,
             // which can modify things like the cookie or env, and then apply the action
             context.Logger?.ModRewriteMatchedRule();
 
-            foreach (var action in Actions)
-            {
-                await action.ApplyActionAsync(context, initMatchRes?.BackReferences, condBackReferences);
-            }
+            return Task.WhenAll(Actions.Select(action => action.ApplyActionAsync(context, initMatchRes?.BackReferences, condResult?.BackReferences)));
```
Tratcher: Does this run the actions in parallel? That's not safe. This method probably still needs async/await.

muratg:
@davidpeden3 closing this for now. Thanks for the contribution so far, and let us know if you'd like to continue with the PR soon.

davidpeden3:
@muratg I thought this one was done. But it was unclear if you guys wanted to merge it in. Can you clarify the status? Thanks.

Tratcher:
The question of usefulness vs perf impact was never resolved. That's on us to resolved. @davidfowl

davidfowl:
Super easy to write a micro benchmark for this so lets do that.

/cc @muratg can you assign somebody to do this?

muratg:
@mikaelm12 could you write a couple micro-benchmarks testing this?

davidpeden3:
@mikaelm12 can you post your results? thanks

mikaelm12:
@davidpeden3 I should have commented when I closed this, sorry.
Unfortunately with all the work we've been doing for 2.0. This didn't make it in 🙁  
We still very much appreciate your contributions though!

jkotalik:
@muratg Could we consider this for 2.1? I can try to get this in based on performance results.

JunTaoLuo:
Changing the `IRule` interface, wouldn't that constitute a breaking change?

jkotalik:
@JunTaoLuo is right. We would either need to add a second interface (IAsyncRule) or something else that is hacky.

PR #215, opened 2017-03-14
Using NullLogger types from Logging.Abstractions

ajaybhargavb:
aspnet/Logging#571
@Eilon


JunTaoLuo:
This is blocking the build so I'm merging it now.

PR #223, opened 2017-04-18
add StringBuilder pool

david-peden-q2:
Add pooling for StringBuilder. This PR addresses the perf concerns noted in a couple of the `Segment` classes. The implementation of `PooledStringBuilder` is borrowed from http://source.roslyn.codeplex.com/#Microsoft.CodeAnalysis.Workspaces/PooledStringBuilder.cs. The implementation of `ObjectPool` is inspired by https://msdn.microsoft.com/en-us/library/ff458671%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396 rather than also borrowing http://source.roslyn.codeplex.com/#Microsoft.CodeAnalysis.Workspaces/ObjectPool%25601.cs. Obviously it would be trivial to switch the implementation out.

I decided to add this as a PR because I ran into needing an additional `StringBuilder` in a PR I'm about to add for adding header support in rewrite rules. Rather than copying and pasting the same perf code in yet another segment, I cleaned it up.

### DIFF
--- a/src/Microsoft.AspNetCore.Rewrite/ObjectPool.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/ObjectPool.cs
@@ -1,0 +1,31 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Concurrent;
+
+namespace Microsoft.AspNetCore.Rewrite
+{
+    public class ObjectPool<T>
+    {
+        private readonly Func<T> _objectFactory;
+        private readonly BlockingCollection<T> _objects;
+
+        public ObjectPool(Func<T> objectFactory, int poolSize)
+        {
+            _objectFactory = objectFactory ?? throw new ArgumentNullException(nameof(objectFactory));
+            _objects = new BlockingCollection<T>(new ConcurrentBag<T>(), poolSize);
+        }
+
+        public T Allocate()
+        {
+            T item;
+            return _objects.TryTake(out item) ? item : _objectFactory();
+        }
+
+        public void Free(T item)
+        {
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNetCore.Rewrite/ObjectPool.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Concurrent;
+
+namespace Microsoft.AspNetCore.Rewrite
+{
+    public class ObjectPool<T>
```
Tratcher: @davidfowl I assume there's a pool implementation available so we don't have to duplicate this?

david-peden-q2: see also:

https://github.com/dotnet/corefx/blob/bffef76f6af208e2042a2f27bc081ee908bb390b/src/System.Reflection.Metadata/src/System/Reflection/Internal/Utilities/PooledStringBuilder.cs

and

https://github.com/dotnet/corefx/blob/bffef76f6af208e2042a2f27bc081ee908bb390b/src/System.Reflection.Metadata/src/System/Reflection/Internal/Utilities/ObjectPool%601.cs

both `internal`

davidfowl:
Can't make performance fixes without before/after performance numbers 😄

davidpeden3:
for sure, so will you guys handle the benchmark for this as you are for https://github.com/aspnet/BasicMiddleware/pull/206?

muratg:
@davidpeden3 thanks for sending out this PR. It became a little stale, so I'll be closing this now. But if you have any measurements (before/after) that you'd like to open up for discussion, please file a bug and let's keep the communication open for a future PR. Thanks.

Issue #190, opened 2016-12-12
RemoteIpAddress and X-Forwarded-For


Tratcher:
_From @JauernigIT on December 12, 2016 12:54_

We struggled for some time now to find a way to reliable find out the client IP address in ASP.NET Core...
- Environment: Windows Server 2008 R2, IIS 7.5, ASP.NET Core 1.0.0 (still)
- Requirement: it should work with direct callers as well as with clients that come over a corporate proxy.

First we activated `UseIISIntegration()` and `app.UseForwardedHeaders(new ForwardedHeadersOptions { ForwardedHeaders = ForwardedHeaders.XForwardedFor })` in the startup phase.

Then we thought that we could use `httpContext.Connection.RemoteIpAddress` for this scenario and yes, it works when we have direct callers - in this case it returns the right client IP... But: unfortunately it doesn't work when the client is behind our corporate proxy. In this case, `Connection.RemoteIpAddress` everytime returns `127.0.0.1`.

Doing some more investigation we found out that the `X-Forwarded-For` request header was correctly set for clients that come over the proxy: `X-Forwarded-For: <client-ip>, 127.0.0.1, <proxy-ip:port>`. Hence, we are able to _manually_ grab the right client IP from the request. But we thought that `httpContext.Connection.RemoteIpAddress` is exactly doing this for us?

_Copied from original issue: aspnet/HttpAbstractions#742_

Tratcher:
_From @khellang on December 12, 2016 16:56_

I think this might be the wrong repo? `ForwardedHeadersMiddleware` belongs in https://github.com/aspnet/BasicMiddleware :smile:

Anyway, have you tried setting the `ForwardLimit` to something other than `1` (the default) on the `ForwardedHeadersOptions`?

Tratcher:
There are a number of other ForwardedHeadersOptions properties to adjust like ForwardLimit and KnownProxies

JauernigIT:
We tried to set `ForwardLimit = null`, as result we got the IP address of the proxy back, but not the _real_ client IP...

How is `httpContext.Connection.RemoteIpAddress` intended to work? Is it evaluating the number of IP addresses in `X-Forwarded-For` request headers depending on `ForwardedHeadersOptions.ForwardLimit`? If so, I thought I would get the real client IP with `ForwardLimit = null`...

khellang:
[...]

PR #226, opened 2017-04-21
Change RequireHeaderSymmetry default to false, improve consistency.

Tratcher:
#190 RequireHeaderSymmetry is a paranoid level of security for forwarded headers that is broken in too many real world scenarios (including Azure). Changing the default to false.

When I changed the default I noticed some inconsistent handling of invalid IPs and cleaned that up a little.

Will follow up by removing the azure hack:
https://github.com/aspnet/IISIntegration/blob/ce3c71b5b3499e83853d9ef1ad59e1e13e65c685/src/Microsoft.AspNetCore.Server.IISIntegration/WebHostBuilderIISExtensions.cs#L61-L66


mikes-gh:
Than ks for this.  This is a good move.  I always have to set it to false as it doesn't work with NGINX or ARR.

PR #227, opened 2017-04-23
Use new Hosting APIs

JunTaoLuo:
https://github.com/aspnet/Hosting/pull/1035


davidfowl:
Agreed with @Tratcher

JunTaoLuo:
We are obsoleting the ConfigureLogging overload that only takes the ILoggerFactory. The one we probably won't deprecate is the UseLoggerFactory overload. I add the context with an explicit name instead _ since this is a sample and it would probably make more sense to be explicit with what that parameter is.

JunTaoLuo:
shouldn't need this anymore

PR #234, opened 2017-05-20
Target .NET Standard 2.0

ajaybhargavb:


JunTaoLuo:
@ajaybhargavb you missed something 😄

pranavkm:
:shipit: when build passes.

pranavkm:
Tests still failing on Windows

JunTaoLuo:
I just revered the change from the previous migration but it looks like there is something different between 46 and 461. I can take a look at it when I get home tonight.

JunTaoLuo:
Looks like ResponseCompression middleware needs to be cross-compiled since gzip stream is not flushable on .NET Framework.

cc @Eilon @DamianEdwards

src/Microsoft.AspNetCore.ResponseCompression/GzipCompressionProvider.cs
```diff
@@ -37,7 +37,13 @@ public bool SupportsFlush
         {
             get
             {
+#if NET461
+                return false;
```
natemcmaster: Is there a way we can detect this without cross-compiling?

pranavkm:
🐑 :it:

src/Microsoft.AspNetCore.ResponseCompression/GzipCompressionProvider.cs
JunTaoLuo: Not unless we want to detect the whether it's .NET Framework or .NET Core at runtime. We wanted to avoid this though.

ajaybhargavb:
Thanks @JunTaoLuo

PR #238, opened 2017-05-25
Fix Redirect bug

mikaelm12:
Fixing a bug in redirect scenarios which causes a redirect loop.

### DIFF
--- a/src/Microsoft.AspNetCore.Rewrite/Internal/RedirectRule.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/Internal/RedirectRule.cs
@@ -62,7 +62,7 @@ namespace Microsoft.AspNetCore.Rewrite.Internal
                     return;
                 }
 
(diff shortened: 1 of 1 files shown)


mikaelm12:
@Tratcher I tried the fix you mentioned with checking the position of `/` vs `://` but that seemed to make other scenarios fail.

src/Microsoft.AspNetCore.Rewrite/Internal/RedirectRule.cs
```diff
@@ -62,7 +62,7 @@ public virtual void ApplyRule(RewriteContext context)
                     return;
                 }
 
-                if (newPath.IndexOf("://", StringComparison.Ordinal) == -1 && newPath[0] != '/')
+                if (!(newPath.StartsWith("http://", StringComparison.Ordinal) || newPath.StartsWith("https://", StringComparison.Ordinal)) && newPath[0] != '/')
```
Tratcher: This case sensitivity will break, it needs to be OrdinalIgnoreCase at least.

PR #243, opened 2017-06-08
Support IOptions<RewriteOptions>

davidfowl:
- Added a test
- Updated the sample to use services

### DIFF
--- a/test/Microsoft.AspNetCore.Rewrite.Tests/MiddlewareTests.cs
+++ b/test/Microsoft.AspNetCore.Rewrite.Tests/MiddlewareTests.cs
@@ -7,6 +7,7 @@ using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.TestHost;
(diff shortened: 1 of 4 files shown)


test/Microsoft.AspNetCore.Rewrite.Tests/MiddlewareTests.cs
```diff
@@ -52,9 +53,31 @@ public class MiddlewareTests
...
+            })
+            .Configure(app =>
+            {
+                app.UseRewriter();
+            });
+            var server = new TestServer(builder);
+
+            var response = await server.CreateClient().GetAsync("foo");
+
+            Assert.Equal("http://example.com/foo", response.Headers.Location.OriginalString);
```
mikaelm12: Check the response status code here too

davidfowl: What response? It's a redirect.

mikaelm12: You can still check the status code that you set in the options above.

PR #253, opened 2017-07-19
Add option to add additional forwarders to ForwardedHeadersMiddleware

YvonneArnoldus:
**Problem**: Not all webservers seem to use the default ForwardedHeaders (X-Forwarded-For, X-Forwarded-Host and X-Forwarded-Proto) but use their own headers. 

**Example**: Azure doesn't the X-Forwarded-Proto header when SSL is used but you need to check the content of X-ARR-SSL to see if https is used.

By making it possible to add additional forwarders to the ForwardedHeadersMiddleware you can use this Middleware for all webservers without the need to create your own custom middleware to catch the webserver that have their own headers.

### DIFF
--- a/samples/HttpOverridesSample/Startup.cs
+++ b/samples/HttpOverridesSample/Startup.cs
@@ -2,13 +2,14 @@ using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.HttpOverrides;
+using Microsoft.Extensions.Logging;
 
 namespace HttpOverridesSample
 {
     public class Startup
     {
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
-        public void Configure(IApplicationBuilder app)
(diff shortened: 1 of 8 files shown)


Tratcher:
Something seems wrong with your branching. Can you rebase your changes on the latest dev?

muratg:
@YvonneArnoldus were you planning to continue to work on this PR?  We will need to review the scenario further and will request some changes if we decide to take this in.

YvonneArnoldus:
@muratg,

I'm done working with the PR unless you need me to make additional changes.

Tratcher:
@YvonneArnoldus The new structure does not provide much value. The abstract Forwarder gives you the HttpContext and no other help, it might as well be a separate middleware. If we're going to help people plug in other forwarders then we should provide a more structured approach for them.

YvonneArnoldus:
@Tratcher I've tried to create a solution that wouldn't break the current implementations. I haven't found a solution to create a more structured approach that wouldn't break the current implementation.

Tratcher:
Thanks for digging into it. Let us know if you come up with anything in the future.

PR #255, opened 2017-07-24
Normalize VersionPrefix

ryanbrandenburg:
Leaving the version of Microsoft.AspNetCore.Buffering alone as I assume there's a reason it's 0.3.0 and not 2.0.0.

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -1,6 +1,6 @@
 ﻿<Project>
   <PropertyGroup>
-    <AspNetCoreVersion>2.0.0-*</AspNetCoreVersion>
+    <AspNetCoreVersion>2.1.0-*</AspNetCoreVersion>
     <InternalAspNetCoreSdkVersion>2.0.1-*</InternalAspNetCoreSdkVersion>
     <MoqVersion>4.7.49</MoqVersion>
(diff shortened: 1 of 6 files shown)


build/dependencies.props
```diff
@@ -1,6 +1,6 @@
 ﻿<Project>
   <PropertyGroup>
-    <AspNetCoreVersion>2.0.0-*</AspNetCoreVersion>
+    <AspNetCoreVersion>$(VersionPrefix)-*</AspNetCoreVersion>
```
natemcmaster: This causes an error:
```
C:\Users\appveyor\.dotnet\x64\sdk\2.0.0-preview3-006841\NuGet.targets(102,5): error : '-*' is not a valid version string.
```
VersionPrefix isn't yet defined at this point in the sample projects

natemcmaster: Simple fix: set this value to

```xml
<AspNetCoreVersion>2.1.0-*</AspNetCoreVersion>
```

ryanbrandenburg: I can swap the import order for dependencies.props and version.props in common.props, so long as this is actually an OK assumption to make.

natemcmaster: Swapping the order doesn't work because samples don't import version.props at all. They only import dependencies.props.

ryanbrandenburg:
:up::date:

Issue #258, opened 2017-08-24
Add first class support for HTTPS though a middleware


javiercn:
We want to have a middleware that does HTTP to HTTPS redirection and possibly other things like HSTS.

javiercn:
This is part of our plan for HTTPS in 2.1

javiercn:
@muratg Can you label this appropiately? There is no needs-design lable here and I can't create one.

muratg:
cc @Tratcher

Tratcher:
There's an implementation available here:
https://github.com/aspnet/BasicMiddleware/blob/793a49fe111d86895f22300297fa70f710459406/src/Microsoft.AspNetCore.Rewrite/Internal/RedirectToHttpsRule.cs

jkotalik:
We could add another extension method that would force HSTS redirection.

javiercn:
I'm more inclined to follow the same approach we do in StaticFiles as @Tratcher suggested to me. We probably should have a meeting to discuss the details.

muratg:
cc @shirhatti @glennc @davidfowl

jkotalik:
@javiercn can you point me to this approach in StaticFiles?

Tratcher:
He's referring to https://github.com/aspnet/StaticFiles/blob/5d2b1000f19dc3e892611f387b46cd519976a9f0/src/Microsoft.AspNetCore.StaticFiles/FileServerExtensions.cs#L91-L101
The features like redirect-to-https and hsts are functionally independent and can be built as separate middleware. Then you add one extension that pulls them all in together.

jkotalik:
@javiercn @Tratcher @shirhatti I have been chunking away at this in my spare time. Here is my stab on the design for the middleware. 

Note, it is required that UseHsts comes before UseRedirectToHttps as UseRedirectToHttps will redirect if the url is not https. See https://tools.ietf.org/html/rfc6797 and https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security for more details on Hsts.

## Extension methods
```csharp
public static IApplicationBuilder UseHttps(this IApplicationBuilder app)
{
}
public static IApplicationBuilder UseHttps(this IApplicationBuilder app, bool enforceHsts)
{
}
public static IApplicationBuilder UseHttps(this IApplicationBuilder app, bool enforceHsts, bool EnforceHttps)
{
}
[...]

PR #264, opened 2017-10-03
Adds Https Redirection and Hsts Middlewares

jkotalik:
Implementation of #258.
Final items:
- Verify which browsers respect the Strict-Transport-Security and check behavior on browser.
- Naming
- Any design/ usability changes

cc @danroth27 @shirhatti @muratg


jkotalik:
Per offline feedback, 
~~HttpsEnforcement~~ -> HttpsPolicy for everything?

src/Microsoft.AspNetCore.HttpsEnforcement/HstsOptions.cs
```diff
@@ -0,0 +1,37 @@
...
+    public class HstsOptions
+    {
+        /// <summary>
+        /// Sets the max-age parameter of the Strict-Transport-Security header.
+        /// </summary>
+        /// <remarks>
+        /// Max-age is required; defaults to 0.
+        /// See: https://tools.ietf.org/html/rfc6797#section-6.1.1
+        /// </remarks>
+        public int MaxAge { get; set; }
```
Tratcher: TimeSpan.
Also, 0 means to disable HSTS, so it can't be the default.
```
   NOTE:  A max-age value of zero (i.e., "max-age=0") signals the UA to
          cease regarding the host as a Known HSTS Host, including the
          includeSubDomains directive (if asserted for that HSTS Host).
```

Tratcher: Though https://tools.ietf.org/html/rfc6797#section-11.2 suggests 0 as the default and making users set it so that it's not using an arbitrary value.
If we require users to set it then we should make it nullable and throw if it's not set. If not then we need to come up with a non-zero default. Talk to @blowdart.

Tratcher: https://tools.ietf.org/html/rfc6797#section-11.2 defines an alternative to a fixed second count where you would define a fixed date and then Max-Age would reflect a count-down to that date. Not sure if that's in wide use or worth implementing. Do other providers offer this variation?

src/Microsoft.AspNetCore.HttpsEnforcement/HstsMiddleware.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// <summary>
+        /// Invoke the middleware.
+        /// </summary>
+        /// <param name="context">The <see cref="HttpContext"/>.</param>
+        /// <returns></returns>
+        public async Task Invoke(HttpContext context)
+        {
+            context.Response.OnStarting(() =>
+            {
+                if (context.Request.Scheme == "https")
```
Tratcher: Request.IsHttps

src/Microsoft.AspNetCore.HttpsEnforcement/HstsOptions.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// <remarks>
+        /// See: https://tools.ietf.org/html/rfc6797#section-6.1.2
+        /// </remarks>
+        public bool IncludeSubDomains { get; set; }
+
+        /// <summary>
+        /// Sets the preload parameter of the Strict-Transport-Security header.
+        /// </summary>
+        /// <remarks>
+        /// Preload is not part of the RFC specification, but is supported by web browsers
```
Tratcher: remove the Chrome reference, just say some web browsers.

src/Microsoft.AspNetCore.HttpsEnforcement/HttpsEnforcementExtensions.cs
```diff
@@ -0,0 +1,66 @@
...
+
+namespace Microsoft.AspNetCore.HttpsEnforcement
+{
+    /// <summary>
+    /// Extension methods for the HttpsEnforcement middleware.
+    /// </summary>
+    public static class HttpsEnforcementExtensions
+    {
+        /// <summary>
+        /// Adds middleware for enforcing HTTPS for all HttpRequests.
```
Tratcher: Including redirecting from HTTP to HTTPS and adding the HSTS header.

src/Microsoft.AspNetCore.HttpsEnforcement/HttpsEnforcementOptions.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        /// <summary>
+        ///  Options for using HSTS
+        /// </summary>
+        public HstsOptions HstsOptions { get; set; } // TODO should this be initialized at all? We could remove the bool if so.
+
+        /// <summary>
+        /// The status code to be used for Url Redirection
+        /// </summary>
+        public int StatusCode { get; set; } = 301; // TODO throw ArgumentOutOfRangeException from UrlRewrite redirect rule?
```
Tratcher: Use the constant

src/Microsoft.AspNetCore.HttpsEnforcement/HttpsEnforcementOptions.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// </summary>
+        public HttpsEnforcementOptions()
+        {
+            HstsOptions = new HstsOptions();
+        }
+
+        /// <summary>
+        /// Whether to use HTTP Strict-Transport-Security (HSTS) on all HTTPS requests.
+        /// </summary>
+        public bool EnforceHsts { get; set; }
```
Tratcher: SendHsts? SetHsts? The HSTS enforcement happens on the client.

samples/HttpsSample/Startup.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        // This method gets called by the runtime. Use this method to add services to the container.
+        // For more information on how to configure your application, visit https://go.microsoft.com/fwlink/?LinkID=398940
+        public void ConfigureServices(IServiceCollection services)
+        {
+        }
+
+        public void Configure(IApplicationBuilder app)
+        {
+            var httpsEnforcementOptions = new HttpsEnforcementOptions();
```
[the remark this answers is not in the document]
javiercn: We should have a helper method for this.
```csharp
services.AddHttpsPolicy(o => o. ...);
```

src/Microsoft.AspNetCore.HttpsEnforcement/HstsExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// </summary>
+        /// <param name="app">The <see cref="IApplicationBuilder"/> instance this method extends.</param>
+        public static IApplicationBuilder UseHsts(this IApplicationBuilder app)
+        {
+            if (app == null)
+            {
+                throw new ArgumentNullException(nameof(app));
+            }
+
+            return app.UseHsts(new HstsOptions());
```
javiercn: HstsOptions here should come from DI.

src/Microsoft.AspNetCore.HttpsEnforcement/HstsMiddleware.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Options;
+using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNetCore.HttpsEnforcement
+{
+    /// <summary>
+    /// Enables Http Strict Transport Security (HSTS)
```
javiercn: nit: `HTTP` (casing)

src/Microsoft.AspNetCore.HttpsEnforcement/HttpsEnforcementExtensions.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// </summary>
+        /// <param name="app">The <see cref="IApplicationBuilder"/> instance this method extends.</param>
+        /// <returns>The <see cref="IApplicationBuilder"/> with HttpsEnforcement.</returns>
+        public static IApplicationBuilder UseHttpsEnforcement(this IApplicationBuilder app)
+        {
+            if (app == null)
+            {
+                throw new ArgumentNullException(nameof(app));
+            }
+            var httpsEnforcementOptions = new HttpsEnforcementOptions();
```
javiercn: This should come from DI

src/Microsoft.AspNetCore.HttpsEnforcement/HstsOptions.cs
jkotalik: I'm in favor of having it nullable and throw if it isn't set. The fixed second count is odd.

Tratcher: Ask the PMs.

src/Microsoft.AspNetCore.HttpsPolicy/HttpsPolicyBuilderExtensions.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// <returns>The <see cref="IApplicationBuilder"/> with HttpsPolicy.</returns>
+        public static IApplicationBuilder UseHttpsPolicy(this IApplicationBuilder app)
+        {
+            if (app == null)
+            {
+                throw new ArgumentNullException(nameof(app));
+            }
+            var httpsEnforcementOptions = new HttpsPolicyOptions();
+
+            return app.UseHttpsPolicy(httpsEnforcementOptions);
```
javiercn: nit: Policy

src/Microsoft.AspNetCore.HttpsPolicy/HttpsPolicyOptions.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        /// <summary>
+        ///  Options for using HSTS
+        /// </summary>
+        public HstsOptions HstsOptions { get; set; } // TODO should this be initialized at all? We could remove the bool if so.
+
+        /// <summary>
+        /// The status code to be used for Url Redirection
+        /// </summary>
+        public int StatusCode { get; set; } = StatusCodes.Status301MovedPermanently; // TODO throw ArgumentOutOfRangeException from UrlRewrite redirect rule?
```
javiercn: It's ok as it is. I don't think we need to over protect against a non redirect status code. I would change the property name and the doc comment to reflect that is a redirect status code. Property would be RedirectStatusCode. What do you think?

src/Microsoft.AspNetCore.HttpsPolicy/HttpsPolicyOptions.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        /// <summary>
+        /// Whether to use HTTP Strict-Transport-Security (HSTS) on all HTTPS requests.
+        /// </summary>
+        public bool SetHsts { get; set; }
+
+        /// <summary>
+        ///  Options for using HSTS
+        /// </summary>
+        public HstsOptions HstsOptions { get; set; } // TODO should this be initialized at all? We could remove the bool if so.
```
javiercn: I'm happy with this being initialized by default (to whatever defaults we choose) and for it being null to have the meaning of SetHsts = false. That way there can't be invalid combinations (like the user initializes the setting and forgets to set the flag or viceversa). I also think we follow this pattern in other places. @Tratcher do you have anything against my proposal?

src/Microsoft.AspNetCore.HttpsPolicy/HttpsPolicyBuilderExtensions.cs
```diff
@@ -65,5 +46,9 @@ public static IApplicationBuilder UseHttpsPolicy(this IApplicationBuilder app, H
 
             return app;
         }
+
+        public static IApplicationBuilder UseHttpsPolicy(this IApplicationBuilder app, bool enableHsts)
+        {
```
muratg: Why empty? Do we need a comment explaining it?

jkotalik: Sorry redesign for this. I need to do a bit of work still!

muratg: Oh, OK, sorry didn't notice the [WIP] in the title.

src/Microsoft.AspNetCore.HttpsPolicy/HstsExtensions.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.AspNetCore.HttpsPolicy
```
Tratcher: Builder namespace

src/Microsoft.AspNetCore.HttpsPolicy/HstsMiddleware.cs
```diff
@@ -0,0 +1,66 @@
...
+                throw new ArgumentNullException(nameof(options));
+            }
+
+            _next = next;
+
+            var hstsOptions = options.Value;
+            var includeSubdomains = hstsOptions.IncludeSubDomains ? IncludeSubDomains : StringSegment.Empty;
+            var preload = hstsOptions.Preload ? Preload : StringSegment.Empty;
+
+            _nameValueHeaderValue = new StringValues($"max-age={hstsOptions.MaxAge}{includeSubdomains}{preload}");
```
Tratcher: _stsValue

src/Microsoft.AspNetCore.HttpsPolicy/HstsMiddleware.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// <param name="context">The <see cref="HttpContext"/>.</param>
+        /// <returns></returns>
+        public async Task Invoke(HttpContext context)
+        {
+
+            if (context.Request.IsHttps)
+            {
+                context.Response.Headers[HeaderNames.StrictTransportSecurity] = _nameValueHeaderValue;
+            }
+            await _next(context);
```
Tratcher: no async/await required

src/Microsoft.AspNetCore.HttpsPolicy/HstsOptions.cs
```diff
@@ -0,0 +1,39 @@
...
+    public class HstsOptions
+    {
+        /// <summary>
+        /// Sets the max-age parameter of the Strict-Transport-Security header.
+        /// </summary>
+        /// <remarks>
+        /// Max-age is required; defaults to 0.
+        /// See: https://tools.ietf.org/html/rfc6797#section-6.1.1
+        /// </remarks>
+        public int MaxAge { get; set; } // TODO set this to a different default in dev/not dev?
```
Tratcher: 0 cannot be the default, that has a specific meaning in the spec. Ask the PMs for a specific value. 1 day? 1 week?
We may not need a separate value for dev as the templates won't have this enabled by default.

src/Microsoft.AspNetCore.HttpsPolicy/HttpsPolicyOptions.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Microsoft.AspNetCore.HttpsPolicy
+{
+    /// <summary>
+    /// Options for the HttpsPolicyMiddleware middleware
+    /// </summary>
+    public class HttpsPolicyOptions
+    {
+        /// <summary>
+        /// Whether to use HTTP Strict-Transport-Security (HSTS) on all HTTPS requests.
```
Tratcher: responses

src/Microsoft.AspNetCore.HttpsPolicy/HttpsPolicyOptions.cs
Tratcher: Yes RedirectStatusCode.

src/Microsoft.AspNetCore.HttpsPolicy/HttpsPolicyOptions.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// </summary>
+        public bool SetHsts { get; set; }
+
+        /// <summary>
+        /// The status code to be used for Url Redirection
+        /// </summary>
+        public int StatusCode { get; set; } = StatusCodes.Status301MovedPermanently; // TODO throw ArgumentOutOfRangeException from UrlRewrite redirect rule?
+
+        /// <summary>
+        /// The TLS port to be added to the redirected URL.
```
Tratcher: The default port (443) will be used if not provided.

test/Microsoft.AspNetCore.HttpsEnforcement.Tests/HttpsPolicyMiddlewareTests.cs
```diff
@@ -0,0 +1,257 @@
...
+        [InlineData(302, null, 0, false, false, "max-age=0", "https://localhost/")]
+        [InlineData(301, 5050, 0, false, false, "max-age=0", "https://localhost:5050/")]
+        [InlineData(301, 443, 0, false, false, "max-age=0", "https://localhost/")]
+        [InlineData(301, 443, 0, true, false, "max-age=0; includeSubDomains", "https://localhost/")]
+        [InlineData(301, 443, 0, false, true, "max-age=0; preload", "https://localhost/")]
+        [InlineData(301, null, 0, true, true, "max-age=0; includeSubDomains; preload", "https://localhost/")]
+        [InlineData(302, 5050, 0, true, true, "max-age=0; includeSubDomains; preload", "https://localhost:5050/")]
+        public async Task SetOptions_AddHstsMiddlewareAndEnableInPolicy(int statusCode, int? tlsPort, int maxAge, bool includeSubDomains, bool preload, string expectedHstsHeader, string expectedUrl)
+        {
+
```
javiercn: nit: remove empty line

test/Microsoft.AspNetCore.HttpsEnforcement.Tests/HttpsPolicyMiddlewareTests.cs
```diff
@@ -0,0 +1,257 @@
...
+                    {
+                        options.IncludeSubDomains = includeSubDomains;
+                        options.MaxAge = maxAge;
+                        options.Preload = preload;
+                    });
+                })
+                .Configure(app =>
+                {
+                    app.UseHttpsPolicy();
+                    app.UseHsts();
```
javiercn: I thought UseHttpsPolicy would add Hsts if needed, why is this line here. Is that not the case?

test/Microsoft.AspNetCore.HttpsEnforcement.Tests/HttpsPolicyMiddlewareTests.cs
```diff
@@ -0,0 +1,257 @@
...
+                });
+
+            var server = new TestServer(builder);
+            var client = server.CreateClient();
+
+            var request = new HttpRequestMessage(HttpMethod.Get, "");
+
+            var response = await client.SendAsync(request);
+
+            // By default, we will not redirect if we only use the Hsts Middleware.
```
javiercn: This comment is confusing, I'm not sure what you are trying to convey here, but it seems out of context

test/Microsoft.AspNetCore.HttpsEnforcement.Tests/HttpsPolicyMiddlewareTests.cs
jkotalik: That is the case. This is just showing that you can still add Hsts again after calling UseHttps.

test/Microsoft.AspNetCore.HttpsEnforcement.Tests/HstsMiddlewareTests.cs
```diff
@@ -0,0 +1,91 @@
...
+                });
+
+            var server = new TestServer(builder);
+            var client = server.CreateClient();
+            client.BaseAddress = new Uri("https://localhost:5050");
+            var request = new HttpRequestMessage(HttpMethod.Get, "");
+
+            var response = await client.SendAsync(request);
+
+            // By default, we will not redirect if we only use the Hsts Middleware.
```
javiercn: You copy pasted this comment below and doesn't make sense there. I don't think this adds a lot of value here, so I would just remove it.

test/Microsoft.AspNetCore.HttpsEnforcement.Tests/HstsMiddlewareTests.cs
```diff
@@ -0,0 +1,91 @@
...
+        public async Task SetOptions_DefaultsSetCorrectly()
+        {
+            var builder = new WebHostBuilder()
+                .UseUrls("https://*:5050")
+                .ConfigureServices(services =>
+                {
+                })
+                .Configure(app =>
+                {
+                    app.UseHsts();
```
javiercn: This is error prone as 0 is not a valid default

src/Microsoft.AspNetCore.HttpsPolicy/HstsOptions.cs
Tratcher: Discussed offline with @javiercn. 0 isn't a valid default, and using a nullable would unnecessarily complicate the user-experience. We just need to come up with a good default. Take it up with @blowdart over email if needs be.

jkotalik: @shirhatti as I know you were opinionated on having 0 as the default.

jkotalik:
We should consider making HttpsPolicy a shared options collection with RedirectToHttpsOptions and HstsOptions. To configure the HttpPolicy, if we require the max-age to be set, users will have to call services.Configure<HstsOptions>.

src/Microsoft.AspNetCore.HttpsEnforcement/HstsOptions.cs
vanillajonathan: The RFC has it "IncludeSubDomains" but one could argue the property should be named `IncludeSubdomains`, I don't know.

jkotalik: I'll stick with the RFC here. 😃

src/Microsoft.AspNetCore.HttpsPolicy/HstsOptions.cs
```diff
@@ -14,13 +14,13 @@ public class HstsOptions
         /// Sets the max-age parameter of the Strict-Transport-Security header.
         /// </summary>
         /// <remarks>
-        /// Max-age is required; defaults to 0.
+        /// Max-age is required; defaults to one month (2592000 seconds).
         /// See: https://tools.ietf.org/html/rfc6797#section-6.1.1
         /// </remarks>
-        public int MaxAge { get; set; }
+        public int MaxAge { get; set; } = 2592000;
```
Tratcher: This should be a TimeSpan so people don't have to do the math themselves.

samples/HttpsPolicySample/Startup.cs
```diff
@@ -0,0 +1,70 @@
...
+namespace HttpsSample
+{
+    public class Startup
+    {
+        // This method gets called by the runtime. Use this method to add services to the container.
+        // For more information on how to configure your application, visit https://go.microsoft.com/fwlink/?LinkID=398940
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.Configure<HttpsRedirectionOptions>(options => {
+                options.RedirectStatusCode = 301;
```
vanillajonathan: Use `StatusCodes.Status301MovedPermanently`

src/Microsoft.AspNetCore.HttpsPolicy/HstsMiddleware.cs
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        /// <summary>
+        /// Invoke the middleware.
+        /// </summary>
+        /// <param name="context">The <see cref="HttpContext"/>.</param>
+        /// <returns></returns>
+        public Task Invoke(HttpContext context)
+        {
+
```
vanillajonathan: We've got a blank line here.

src/Microsoft.AspNetCore.HttpsPolicy/HstsMiddleware.cs
```diff
@@ -0,0 +1,66 @@
...
+            if (context.Request.IsHttps)
+            {
+                context.Response.Headers[HeaderNames.StrictTransportSecurity] = _strictTransportSecurityValue;
+            }
+            return  _next(context);
```
vanillajonathan: Should have a blank line before the return statement.

samples/HttpsPolicySample/Startup.cs
```diff
@@ -0,0 +1,70 @@
...
+
+namespace HttpsSample
+{
+    public class Startup
+    {
+        // This method gets called by the runtime. Use this method to add services to the container.
+        // For more information on how to configure your application, visit https://go.microsoft.com/fwlink/?LinkID=398940
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.Configure<HttpsRedirectionOptions>(options => {
```
javiercn: Update the samples to use the helper methods

jkotalik:
@javiercn @Tratcher We need to talk about how configuration options will be propagated to the middleware. Aka does this middleware need to do anything different or will configuration call app.UseHsts() or app.HttpsRedirection().

src/Microsoft.AspNetCore.HttpsPolicy/HstsMiddleware.cs
```diff
@@ -0,0 +1,66 @@
...
+                throw new ArgumentNullException(nameof(options));
+            }
+
+            _next = next;
+
+            var hstsOptions = options.Value;
+            var includeSubdomains = hstsOptions.IncludeSubDomains ? IncludeSubDomains : StringSegment.Empty;
+            var preload = hstsOptions.Preload ? Preload : StringSegment.Empty;
+
+            _strictTransportSecurityValue = new StringValues($"max-age={hstsOptions.MaxAge.TotalSeconds}{includeSubdomains}{preload}");
```
Tratcher: TotalSeconds needs formatting. Compare to https://github.com/aspnet/Security/blob/bd07f8b683ce793490d108b2310fa6112953d172/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs#L361-L362

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionOptions.cs
```diff
@@ -0,0 +1,27 @@
...
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.HttpsPolicy
+{
+    /// <summary>
+    /// Options for the HttpsRedirection middleware
+    /// </summary>
+    public class HttpsRedirectionOptions
+    {
+
```
Tratcher: extra line

src/Microsoft.AspNetCore.HttpsPolicy/Microsoft.AspNetCore.HttpsPolicy.csproj
```diff
@@ -0,0 +1,19 @@
...
+    <Description>
+      ASP.NET Core basic middleware for supporting HTTPS Redirection and HTTP Strict-Transport-Security.
+    </Description>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <NoWarn>$(NoWarn);CS1591</NoWarn>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <PackageTags>aspnetcore;https;hsts</PackageTags>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Http.Extensions" />
```
Tratcher: What is Http.Extensions needed for?

Tratcher:
The current Options usage looks fine.

src/Microsoft.AspNetCore.HttpsPolicy/HstsMiddleware.cs
jkotalik: I also need to check for if the MaxAge has value then. Default to the same default in HstsOptions if no value?

jkotalik: nvrm realized I realized that security has max age as nullable.

test/Microsoft.AspNetCore.HttpsEnforcement.Tests/HstsMiddlewareTests.cs
```diff
@@ -0,0 +1,128 @@
...
+        [InlineData(50000, false, true, "max-age=50000; preload")]
+        [InlineData(0, true, true, "max-age=0; includeSubDomains; preload")]
+        [InlineData(50000, true, true, "max-age=50000; includeSubDomains; preload")]
+        public async Task SetOptionsThroughConfigure_SetsHeaderCorrectly(int maxAge, bool includeSubDomains, bool preload, string expected)
+        {
+            var builder = new WebHostBuilder()
+                .UseUrls("https://*:5050")
+                .ConfigureServices(services =>
+                {
+                    services.Configure<HstsOptions>(options => {
```
javiercn: Fix the usages on the tests to use the service collection extensions

test/Microsoft.AspNetCore.HttpsEnforcement.Tests/HttpsPolicyTests.cs
```diff
@@ -0,0 +1,76 @@
...
+        [InlineData(301, null, 2592000, true, true, "max-age=2592000; includeSubDomains; preload", "https://localhost/")]
+        [InlineData(302, 5050, 2592000, true, true, "max-age=2592000; includeSubDomains; preload", "https://localhost:5050/")]
+        public async Task SetsBothHstsAndHttpsRedirection_RedirectOnFirstRequest_HstsOnSecondRequest(int statusCode, int? tlsPort, int maxAge, bool includeSubDomains, bool preload, string expectedHstsHeader, string expectedUrl)
+        {
+
+            var builder = new WebHostBuilder()
+                .UseUrls("https://*:5050", "http://*:5050")
+                .ConfigureServices(services =>
+                {
+                    services.Configure<HttpsRedirectionOptions>(options =>
```
javiercn: Here too

test/Microsoft.AspNetCore.HttpsEnforcement.Tests/HttpsRedirectionMiddlewareTests.cs
```diff
@@ -0,0 +1,125 @@
...
+        [InlineData(308, null, "https://localhost/")]
+        [InlineData(301, 5050, "https://localhost:5050/")]
+        [InlineData(301, 443, "https://localhost/")]
+        public async Task SetOptions_SetStatusCodeTlsPort(int statusCode, int? tlsPort, string expected)
+        {
+
+            var builder = new WebHostBuilder()
+                .ConfigureServices(services =>
+                {
+                    services.Configure<HttpsRedirectionOptions>(options =>
```
javiercn: Here

test/Microsoft.AspNetCore.HttpsEnforcement.Tests/HstsMiddlewareTests.cs
jkotalik: There are two sets of tests, one that uses Configure<HstsOptions> and AddHsts(). I can remove these tests if you would like 😄

PR #266, opened 2017-10-28
Adds ASPNETCORE_SSL_PORT as a way to get port

jkotalik:
Also changes the name from TLSPort to SSLPort. 
For our HTTPS story.

### DIFF
--- a/samples/HttpsPolicySample/Startup.cs
+++ b/samples/HttpsPolicySample/Startup.cs
@@ -21,7 +21,7 @@ namespace HttpsSample
             services.AddHttpsRedirection(options =>
             {
                 options.RedirectStatusCode = StatusCodes.Status301MovedPermanently;
-                options.SSLPort = 5001;
+                options.HttpsPort = 5001;
             });
 
             services.AddHsts(options =>
--- a/samples/HttpsPolicySample/Startup.cs
+++ b/samples/HttpsPolicySample/Startup.cs
@@ -21,7 +21,7 @@ namespace HttpsSample
             services.AddHttpsRedirection(options =>
             {
                 options.RedirectStatusCode = StatusCodes.Status301MovedPermanently;
-                options.TlsPort = 5001;
+                options.SSLPort = 5001;
             });
 
             services.AddHsts(options =>
--- a/src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
@@ -33,24 +33,24 @@ namespace Microsoft.AspNetCore.Builder
             var options = app.ApplicationServices.GetRequiredService<IOptions<HttpsRedirectionOptions>>().Value;
 
             // The tls port set in options will have priority over the one in configuration.
-            var tlsPort = options.SSLPort;
-            if (tlsPort == null)
+            var httpsPort = options.HttpsPort;
+            if (httpsPort == null)
             {
                 // Only read configuration if there is no tlsPort
                 var config = app.ApplicationServices.GetRequiredService<IConfiguration>();
-                var configTlsPort = config["SSL_PORT"];
+                var configHttpsPort = config["HTTPS_PORT"];
                 // If the string isn't empty, try to parse it.
-                if (!string.IsNullOrEmpty(configTlsPort)
-                    && int.TryParse(configTlsPort, out var intTlsPort))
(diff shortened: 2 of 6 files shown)


samples/HttpsPolicySample/Startup.cs
```diff
@@ -21,7 +21,7 @@ public void ConfigureServices(IServiceCollection services)
             services.AddHttpsRedirection(options =>
             {
                 options.RedirectStatusCode = StatusCodes.Status301MovedPermanently;
-                options.TlsPort = 5001;
+                options.SSLPort = 5001;
```
javiercn: The correct term is Tls, please revert this change and avoid using SSL

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
```diff
@@ -31,10 +32,25 @@ public static IApplicationBuilder UseHttpsRedirection(this IApplicationBuilder a
 
             var options = app.ApplicationServices.GetRequiredService<IOptions<HttpsRedirectionOptions>>().Value;
 
+            // The tls port set in options will have priority over the one in configuration.
+            var tlsPort = options.SSLPort;
+            if (tlsPort == null)
+            {
+                // Only read configuration if there is no tlsPort
+                var config = app.ApplicationServices.GetRequiredService<IConfiguration>();
+                var configTlsPort = config["SSL_PORT"];
```
javiercn: @tratcher Should this be TLS_PORT or TLSPORT or HTTPSPORT or HTTPS_PORT

samples/HttpsPolicySample/Startup.cs
Tratcher: The same property is called SslPort at least two other places in the stack. Consistency is easier to understand. We don't actually know what protocol the underlying server is using, and most people don't know the difference anyways.

Tratcher: https://github.com/search?l=C%23&q=org%3Aaspnet+SslPort&type=Code
SslPort: launchsettings.json, MvcOptions, RequireHttpsAttribute, AddRedirectToHttps.

javiercn: Two wrongs don’t make a right. Should we just call it https port moving forward? That way we use the scheme instead of the protocol.

MvcOptions and RequireHttpsAttribute are not recommended anymore. The port on mvcoptions is for development mostly.

Tratcher: These are all primarily for development.

It doesn't merit a long debate. I appeal to @Eilon as the namer of arbitrary things.

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
jkotalik: I'm okay with changing it to HTTPS_PORT, but I agree with @Tratcher for SSL > TLS.

samples/HttpsPolicySample/Startup.cs
Eilon: If we're creating new names, I think they should use the correct terminology, i.e. `HTTPS` and not `SSL`. Even though `TLS` might be correct right now, who knows what the future will hold. The term `HTTPS` should hopefully outlive all of those.

Eilon: (Unless, that is, something is actually referring to a specific lower-level protocol. But that doesn't appear to be the case here.)

jkotalik:
@Tratcher is there a place where we store environment variable strings in our stack? If not, we should consider adding one 😄

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
```diff
@@ -31,10 +32,25 @@ public static IApplicationBuilder UseHttpsRedirection(this IApplicationBuilder a
 
             var options = app.ApplicationServices.GetRequiredService<IOptions<HttpsRedirectionOptions>>().Value;
 
+            // The tls port set in options will have priority over the one in configuration.
+            var httpsPort = options.HttpsPort;
+            if (httpsPort == null)
+            {
+                // Only read configuration if there is no tlsPort
```
Tratcher: httpsPort

Tratcher:
No, there's no central env var constants list. It's too component specific.

Issue #182, opened 2016-12-02
Add more GZIP compressable MIME types


RehanSaeed:
I raised the following PR https://github.com/aspnet/BasicMiddleware/pull/181#issuecomment-264494906. This is a justifications for each MIME type to be added:

- `application/rss+xml`, `application/atom+xml` and `image/svg+xml` These are all just special dialects of XML, they are all web related and are fairly common (especially SVG).
- `application/vnd.ms-fontobject`, `application/x-font-ttf`, `font/otf` - TrueType fonts (ttf) and OpenType fonts (otf) can be safely be compressed for smaller size. The answer is unclear when it comes to Embedded OpenType (eot) fonts. It’s possible to compress EOT font files as part of the file creation process, in which case gzipping them is unnecessary. However, to generate compressed EOT font files requires a special tool and an extra step, so it’s likely that most of them are not pre-compressed. However, [Google](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/webfont-optimization#reducing_font_size_with_compression) recommends compressing EOT fonts.
- `image/x-icon` - Finding references for this is more difficult and I can only provide my own anecdotal evidence and the fact that the [HTML 5 Boilerplate Server Configs](https://github.com/h5bp/server-configs/) project also includes it for NGINX and IIS.

Tratcher:
@glennc

glennc:
Adding to 2.0 to prototype API changes for adding/removing mimetypes then re-evaluating this based on easier usage instead of modifying defaults.

muratg:
@glennc Do you have any prototype API surface that you'd like to make adding/removing mimetypes easy?

glennc:
I did have, but that was a long time ago. Let me go see.

glennc:
At this point I don't think we will get to this one in 2.0. Moving to the punt bucket, if there is time we can look at it. But more likely it will slip to a later release.

muratg:
@glennc move the milestone as well?

muratg:
Triage: 
 1. Verify if the extensibility points are sufficient, so folks can add any MIME type they care about
 2. Create a blessed list of MIME types and include them by default

muratg:
@shirhatti I'm assuming you'll move this to 3.0?

PR #270, opened 2017-11-05
Add image/svg+xml MIME type to be compressed

buvinghausen:
Your default MIME types to compress list is pretty comprehensive but it's missing the SVG mime type which is ultimately XML.

Even the example posted about adding Brotli support adds it:

https://blogs.msdn.microsoft.com/dotnet/2017/07/27/introducing-support-for-brotli-compression/


Tratcher:
Closing this as a duplicate for now. We can revisit it when we get https://github.com/aspnet/BasicMiddleware/issues/182 resolved.

Issue #275, opened 2017-11-29
HttpsRedirection should look at IServerAddressesFeature for HTTPS ports


jkotalik:
If Kestrel is listening on any HTTPS uri, the HttpsRedirection middleware should pick the first one from the IServerAddressesFeature.
cc/ @Tratcher @danroth27

Tratcher:
But throw if there's more than one unique https port.

PR #276, opened 2017-11-29
Read from IServerAddressesFeature for HTTPS port

jkotalik:
For #275 
Few notes:
- The middleware requires that the IServerAddressesFeature is implemented now (shouldn't be a problem, just something to note for tests)
- I needed to create a new HttpsMiddleware (rather than use Urlrewrite) as we need to inject the IServerAddressFeature and evaluate it on first request.

### DIFF
--- a/src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
@@ -38,19 +38,16 @@ namespace Microsoft.AspNetCore.Builder
             // 2. HTTPS_PORT environment variable
             // 3. IServerAddressesFeature
             // 4. 443 (or not set.)
-
-
             var httpsPort = 0;
             if (options.HttpsPort == null)
             {
-                if (TryGetHttpsPortFromEnvironmentVariable(app, out httpsPort)
-                    || TryGetHttpsPortFromIServerAddressesFeature(app, out httpsPort))
+                if (TryGetHttpsPortFromEnvironmentVariable(app, out httpsPort))
                 {
                     options.HttpsPort = httpsPort;
                 }
             }
 
-            app.UseMiddleware<HttpsRedirectionMiddleware>();
+            app.UseMiddleware<HttpsRedirectionMiddleware>(app.ServerFeatures.Get<IServerAddressesFeature>());
 
             return app;
         }
@@ -68,14 +65,5 @@ namespace Microsoft.AspNetCore.Builder
             return false;
         }
 
-        private static bool TryGetHttpsPortFromIServerAddressesFeature(IApplicationBuilder app, out int httpsPort)
-        {
-            httpsPort = 0;
-            var addressFeature = app.ServerFeatures.Get<IServerAddressesFeature>();
-            foreach (var address in addressFeature.Addresses)
-            {
-            }
-            return false;
-        }
     }
 }
--- a/src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
@@ -4,7 +4,6 @@
 using System;
 using Microsoft.AspNetCore.Hosting.Server.Features;
 using Microsoft.AspNetCore.HttpsPolicy;
-using Microsoft.AspNetCore.Rewrite;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
@@ -32,25 +31,9 @@ namespace Microsoft.AspNetCore.Builder
             }
 
             var options = app.ApplicationServices.GetRequiredService<IOptions<HttpsRedirectionOptions>>().Value;
+            var config = app.ApplicationServices.GetRequiredService<IConfiguration>();
 
-            // Order for finding the HTTPS port:
-            // 1. Set in the HttpsRedirectionOptions
-            // 2. HTTPS_PORT environment variable
-            // 3. IServerAddressesFeature (checked in HttpsRedirectionMiddleware.Invoke
-            // 4. 443 (or not set)
-            if (options.HttpsPort == null)
-            {
-                var config = app.ApplicationServices.GetRequiredService<IConfiguration>();
-                var configHttpsPort = config["HTTPS_PORT"];
-                // If the string isn't empty, try to parse it.
-                if (!string.IsNullOrEmpty(configHttpsPort)
-                                    && int.TryParse(configHttpsPort, out var intHttpsPort))
-                {
-                        options.HttpsPort = intHttpsPort;
-                }
-            }
-
-            app.UseMiddleware<HttpsRedirectionMiddleware>(app.ServerFeatures.Get<IServerAddressesFeature>());
+            app.UseMiddleware<HttpsRedirectionMiddleware>(app.ServerFeatures.Get<IServerAddressesFeature>(), config);
 
             return app;
         }
--- a/src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
@@ -38,19 +38,15 @@ namespace Microsoft.AspNetCore.Builder
             // 2. HTTPS_PORT environment variable
             // 3. IServerAddressesFeature (checked in HttpsRedirectionMiddleware.Invoke
             // 4. 443 (or not set)
-            var httpsPort = 0;
             if (options.HttpsPort == null)
             {
-                httpsPort = 0;
                 var config = app.ApplicationServices.GetRequiredService<IConfiguration>();
                 var configHttpsPort = config["HTTPS_PORT"];
                 // If the string isn't empty, try to parse it.
-                if (!string.IsNullOrEmpty(configHttpsPort))
+                if (!string.IsNullOrEmpty(configHttpsPort)
+                                    && int.TryParse(configHttpsPort, out var intHttpsPort))
                 {
-                    if (int.TryParse(configHttpsPort, out httpsPort))
-                    {
-                        options.HttpsPort = httpsPort;
-                    }
+                        options.HttpsPort = intHttpsPort;
                 }
             }
 
--- a/src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
@@ -31,9 +31,8 @@ namespace Microsoft.AspNetCore.Builder
             }
 
             var options = app.ApplicationServices.GetRequiredService<IOptions<HttpsRedirectionOptions>>().Value;
-            var config = app.ApplicationServices.GetRequiredService<IConfiguration>();
 
-            app.UseMiddleware<HttpsRedirectionMiddleware>(app.ServerFeatures.Get<IServerAddressesFeature>(), config);
+            app.UseMiddleware<HttpsRedirectionMiddleware>(app.ServerFeatures.Get<IServerAddressesFeature>());
 
             return app;
         }
--- a/src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using Microsoft.AspNetCore.Hosting.Server.Features;
 using Microsoft.AspNetCore.HttpsPolicy;
 using Microsoft.AspNetCore.Rewrite;
 using Microsoft.Extensions.Configuration;
@@ -32,29 +33,49 @@ namespace Microsoft.AspNetCore.Builder
 
             var options = app.ApplicationServices.GetRequiredService<IOptions<HttpsRedirectionOptions>>().Value;
 
-            // The tls port set in options will have priority over the one in configuration.
-            var httpsPort = options.HttpsPort;
-            if (httpsPort == null)
+            // Order for finding the HTTPS port:
+            // 1. Set in the HttpsRedirectionOptions
+            // 2. HTTPS_PORT environment variable
+            // 3. IServerAddressesFeature
+            // 4. 443 (or not set.)
+
+
+            var httpsPort = 0;
+            if (options.HttpsPort == null)
             {
-                // Only read configuration if there is no httpsPort
-                var config = app.ApplicationServices.GetRequiredService<IConfiguration>();
-                var configHttpsPort = config["HTTPS_PORT"];
-                // If the string isn't empty, try to parse it.
-                if (!string.IsNullOrEmpty(configHttpsPort)
-                    && int.TryParse(configHttpsPort, out var intHttpsPort))
+                if (TryGetHttpsPortFromEnvironmentVariable(app, out httpsPort)
+                    || TryGetHttpsPortFromIServerAddressesFeature(app, out httpsPort))
                 {
-                    httpsPort = intHttpsPort;
+                    options.HttpsPort = httpsPort;
                 }
             }
 
-            var rewriteOptions = new RewriteOptions();
-            rewriteOptions.AddRedirectToHttps(
-                options.RedirectStatusCode,
-                httpsPort);
-
-            app.UseRewriter(rewriteOptions);
+            app.UseMiddleware<HttpsRedirectionMiddleware>();
 
             return app;
+        }
+
+        private static bool TryGetHttpsPortFromEnvironmentVariable(IApplicationBuilder app, out int httpsPort)
+        {
+            httpsPort = 0;
+            var config = app.ApplicationServices.GetRequiredService<IConfiguration>();
+            var configHttpsPort = config["HTTPS_PORT"];
+            // If the string isn't empty, try to parse it.
+            if (!string.IsNullOrEmpty(configHttpsPort))
+            {
+                return int.TryParse(configHttpsPort, out httpsPort);
+            }
+            return false;
+        }
+
+        private static bool TryGetHttpsPortFromIServerAddressesFeature(IApplicationBuilder app, out int httpsPort)
+        {
+            httpsPort = 0;
+            var addressFeature = app.ServerFeatures.Get<IServerAddressesFeature>();
+            foreach (var address in addressFeature.Addresses)
+            {
+            }
+            return false;
         }
     }
 }
--- a/src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
@@ -36,14 +36,21 @@ namespace Microsoft.AspNetCore.Builder
             // Order for finding the HTTPS port:
             // 1. Set in the HttpsRedirectionOptions
             // 2. HTTPS_PORT environment variable
-            // 3. IServerAddressesFeature
-            // 4. 443 (or not set.)
+            // 3. IServerAddressesFeature (checked in HttpsRedirectionMiddleware.Invoke
+            // 4. 443 (or not set)
             var httpsPort = 0;
             if (options.HttpsPort == null)
             {
-                if (TryGetHttpsPortFromEnvironmentVariable(app, out httpsPort))
+                httpsPort = 0;
+                var config = app.ApplicationServices.GetRequiredService<IConfiguration>();
+                var configHttpsPort = config["HTTPS_PORT"];
+                // If the string isn't empty, try to parse it.
+                if (!string.IsNullOrEmpty(configHttpsPort))
                 {
-                    options.HttpsPort = httpsPort;
+                    if (int.TryParse(configHttpsPort, out httpsPort))
+                    {
+                        options.HttpsPort = httpsPort;
+                    }
                 }
             }
 
@@ -51,19 +58,5 @@ namespace Microsoft.AspNetCore.Builder
 
             return app;
         }
-
-        private static bool TryGetHttpsPortFromEnvironmentVariable(IApplicationBuilder app, out int httpsPort)
-        {
-            httpsPort = 0;
-            var config = app.ApplicationServices.GetRequiredService<IConfiguration>();
-            var configHttpsPort = config["HTTPS_PORT"];
-            // If the string isn't empty, try to parse it.
-            if (!string.IsNullOrEmpty(configHttpsPort))
-            {
-                return int.TryParse(configHttpsPort, out httpsPort);
-            }
-            return false;
-        }
-
     }
 }
--- a/src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
+++ b/src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
@@ -2,8 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
-using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Hosting.Server.Features;
 using Microsoft.AspNetCore.Http;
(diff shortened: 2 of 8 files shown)


src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
```diff
@@ -0,0 +1,114 @@
...
+            if (options == null)
+            {
+                throw new ArgumentNullException(nameof(options));
+            }
+            _next = next;
+
+            var httpsRedirectionOptions = options.Value;
+            _httpsPort = httpsRedirectionOptions.HttpsPort;
+            _statusCode = httpsRedirectionOptions.RedirectStatusCode;
+            _serverAddressesFeature = serverAddressesFeature;
```
Tratcher: Null check?

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        /// <summary>
+        /// Initializes the HttpsRedirectionMiddleware
+        /// </summary>
+        /// <param name="next"></param>
+        /// <param name="serverAddressesFeature">The</param>
+        /// <param name="options"></param>
+        public HttpsRedirectionMiddleware(RequestDelegate next, IServerAddressesFeature serverAddressesFeature, IOptions<HttpsRedirectionOptions> options)
+        {
+            if (next == null)
```
Tratcher: inline?

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
```diff
@@ -0,0 +1,114 @@
...
+            return _next(context);
+        }
+
+        private void CheckAddressesFeatureForHttpsPorts()
+        {
+            // The IServerAddressesFeature will not be ready until the middleware is Invoked,
+            int? httpsPort = null;
+            foreach (var address in _serverAddressesFeature.Addresses)
+            {
+                if (Uri.TryCreate(address, UriKind.Absolute, out var uri))
```
Tratcher: Parsing is going to be one of the messier parts of this. You have to duplicate Kestrel's parser. Uri doesn't understand "https://+:8080/"
https://github.com/aspnet/KestrelHttpServer/blob/01e2c881317ef3473b313a01eb194f6f418b11b6/src/Kestrel.Core/ServerAddress.cs#L65
@halter73

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
```diff
@@ -0,0 +1,114 @@
...
+                {
+                    if (uri.Scheme == "https")
+                    {
+                        // If we find multiple https ports specified, throw
+                        if (httpsPort != null)
```
Tratcher: allow multiple if the ports match. E.g. if the ports are all X then it will be OK.

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        /// <summary>
+        /// Invokes the HttpsRedirectionMiddleware
+        /// </summary>
+        /// <param name="context"></param>
+        /// <returns></returns>
+        public Task Invoke(HttpContext context)
+        {
+            if (!context.Request.IsHttps)
```
Tratcher: Invert condition to reduce nesting?

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
```diff
@@ -0,0 +1,114 @@
...
+            {
+                if (!_evaluatedServerAddressesFeature && _httpsPort == null)
+                {
+                    CheckAddressesFeatureForHttpsPorts();
+                }
+                _evaluatedServerAddressesFeature = true;
```
Tratcher: Why do you need _evaluatedServerAddressesFeature ? Have CheckAddressesFeatureForHttpsPorts always set _httpsPort, even if it's only 443. (except in the exception case)

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
```diff
@@ -0,0 +1,114 @@
...
+        /// <summary>
+        /// Invokes the HttpsRedirectionMiddleware
+        /// </summary>
+        /// <param name="context"></param>
+        /// <returns></returns>
+        public Task Invoke(HttpContext context)
+        {
+            if (!context.Request.IsHttps)
+            {
+                if (!_evaluatedServerAddressesFeature && _httpsPort == null)
```
Tratcher: !_httpsPort.HasValue

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
```diff
@@ -0,0 +1,114 @@
...
+                {
+                    CheckAddressesFeatureForHttpsPorts();
+                }
+                _evaluatedServerAddressesFeature = true;
+
+                var host = context.Request.Host;
+                if (_httpsPort.HasValue && _httpsPort.Value > 0)
```
Tratcher: It should always have a value by the time you get here.
how would it be set to 0? Shouldn't that be validated earlier?

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
```diff
@@ -0,0 +1,114 @@
...
+                {
+                    host = new HostString(host.Host, _httpsPort.Value);
+                }
+                else
+                {
+                    host = new HostString(host.Host);
+                }
+
+                var request = context.Request;
+                var redirectUrl = UriHelper.BuildAbsolute("https",
```
javiercn: var redirectUrl = UriHelper.BuildAbsolute(
    "https", 
    host,
    request.PathBase,
    request.Path,
    request.QueryString);

Tratcher: ? are you referring to the indenting?

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
javiercn: Can we share the code for those things?

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
jkotalik: We should include indentation style here if it isn't already: https://github.com/aspnet/Home/wiki/Engineering-guidelines

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
jkotalik: I copied this from UrlRewrite, probably was wrong there.

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
jkotalik: I only want to read from the IServerAddressesFeature once, right? It will set the value in the middleware and not need to do it again.

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
```diff
@@ -32,27 +33,24 @@ public static IApplicationBuilder UseHttpsRedirection(this IApplicationBuilder a
 
             var options = app.ApplicationServices.GetRequiredService<IOptions<HttpsRedirectionOptions>>().Value;
 
-            // The tls port set in options will have priority over the one in configuration.
-            var httpsPort = options.HttpsPort;
-            if (httpsPort == null)
+            // Order for finding the HTTPS port:
+            // 1. Set in the HttpsRedirectionOptions
+            // 2. HTTPS_PORT environment variable
```
javiercn: This will mean that any reverse proxy will have to setup this environment variable, even when listening to 443 on the external address

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
javiercn: @Tratcher Yes

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
```diff
@@ -32,27 +33,24 @@ public static IApplicationBuilder UseHttpsRedirection(this IApplicationBuilder a
 
             var options = app.ApplicationServices.GetRequiredService<IOptions<HttpsRedirectionOptions>>().Value;
```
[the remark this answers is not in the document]
javiercn: Can we actually do it on Invoke? We need to wait until we have a request to figure out the correct HTTPS port to redirect to, so we might just do everything in one place.

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
javiercn: Actually, I had an idea for this. Can we make the forwarding headers middleware play along with this one (by putting a feature on the request) and have this middleware not use the server addresses feature if the request was forwarded? That way the logic will be:

// Order for finding the HTTPS port:
// 1. Set in the HttpsRedirectionOptions
// 2. HTTPS_PORT environment variable
// 3. if the request was not forwarded, IServerAddressesFeature
// 4. 443 (or not set)

Tratcher: That's way too many dominoes. Let's not make this any more complicated than it needs to be.

Reading IServerAddressesFeature is only a problem if the local server has HTTPS it's not on the same port as the proxy. The average reverse proxy uses HTTP to communicate with the back-end.

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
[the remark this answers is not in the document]
jkotalik: To clarify, the HTTPS_PORT env variable may not be defined until the first request? Not sure why that is.

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
jkotalik: Also, httpsPort not having a value implies 443 (empty)?

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
jkotalik: I think I can just use Kestrel's ServerAddress.FromUri function?

jkotalik:
I believe I addressed most of the feedback.

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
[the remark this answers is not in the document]
javiercn: HTTPS_PORT should definetely be defined when the middleware gets instantiated. The only thing undefined until the first request is IServerAddressesFeature. @Tratcher can correct me

src/Microsoft.AspNetCore.HttpsPolicy/Microsoft.AspNetCore.HttpsPolicy.csproj
```diff
@@ -11,6 +11,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\..\src\Microsoft.AspNetCore.Rewrite\Microsoft.AspNetCore.Rewrite.csproj" />
+    <PackageReference Include="Microsoft.AspNetCore.Server.Kestrel.Core" Version="$(MicrosoftAspNetCoreServerKestrelCorePackageVersion)" />
```
javiercn: :s We don't want to do this. You need to extract the logic into a sources package

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
```diff
@@ -0,0 +1,134 @@
...
+                                && int.TryParse(configHttpsPort, out var intHttpsPort))
+            {
+                _httpsPort = intHttpsPort;
+                return;
+            }
+
+            int? httpsPort = null;
+            foreach (var address in _serverAddressesFeature.Addresses)
+            {
+                var serverAddress = ServerAddress.FromUrl(address);
```
javiercn: This needs to come from a sources package. We don't want this middleware to depend on Kestrel. /cc: @Tratcher

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
```diff
@@ -31,28 +31,9 @@ public static IApplicationBuilder UseHttpsRedirection(this IApplicationBuilder a
...
-                }
-            }
-
-            var rewriteOptions = new RewriteOptions();
-            rewriteOptions.AddRedirectToHttps(
-                options.RedirectStatusCode,
-                httpsPort);
-
-            app.UseRewriter(rewriteOptions);
+            app.UseMiddleware<HttpsRedirectionMiddleware>(app.ServerFeatures.Get<IServerAddressesFeature>(), config);
```
javiercn: You don't have to pass in the config here. If declared on the constructor it will just get injected.

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
```diff
@@ -0,0 +1,134 @@
...
+            // 4. 443 (or not set)
+
+            if (_httpsPort != null)
+            {
+                return;
+            }
+
+            var configHttpsPort = _config["HTTPS_PORT"];
+            if (!string.IsNullOrEmpty(configHttpsPort)
+                                && int.TryParse(configHttpsPort, out var intHttpsPort))
```
javiercn: Indentation/style

if (!string.IsNullOrEmpty(configHttpsPort) &&
    int.TryParse(configHttpsPort, out var intHttpsPort))

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
```diff
@@ -0,0 +1,134 @@
...
+                return;
+            }
+
+            int? httpsPort = null;
+            foreach (var address in _serverAddressesFeature.Addresses)
+            {
+                var serverAddress = ServerAddress.FromUrl(address);
+                if (serverAddress.Scheme == "https")
+                {
+
```
javiercn: nit: remove extra line

test/Microsoft.AspNetCore.HttpsEnforcement.Tests/HttpsRedirectionMiddlewareTests.cs
```diff
@@ -123,13 +131,16 @@ public class HttpsRedirectionMiddlewareTests
...
-        [InlineData(null, null, "https://localhost/")]
-        [InlineData(null, "5000", "https://localhost:5000/")]
-        [InlineData(null, "443", "https://localhost/")]
-        [InlineData(443, "5000", "https://localhost/")]
-        [InlineData(4000, "5000", "https://localhost:4000/")]
-        [InlineData(5000, null, "https://localhost:5000/")]
-        public async Task SetHttpsPortEnvironmentVariable_ReturnsCorrectStatusCodeOnResponse(int? optionsHttpsPort, string configHttpsPort, string expectedUrl)
+        [InlineData(null, null, null, "https://localhost/")]
+        [InlineData(null, null, "https://localhost:4444/", "https://localhost:4444/")]
+        [InlineData(null, null, "https://localhost:443/", "https://localhost/")]
```
javiercn: Add a test where this url has an http scheme

test/Microsoft.AspNetCore.HttpsEnforcement.Tests/HttpsRedirectionMiddlewareTests.cs
```diff
@@ -157,5 +178,71 @@ public class HttpsRedirectionMiddlewareTests
...
+
+            var request = new HttpRequestMessage(HttpMethod.Get, "");
+
+            var response = await client.SendAsync(request);
+
+            Assert.Equal("https://localhost:5050/", response.Headers.Location.ToString());
+        }
+
+        [Fact]
+        public async Task SetServerAddressesFeature_MultipleHttpsAddresses_ThrowInMiddleware()
```
javiercn: Add a test that has multiple HTTPS urls but all of them pointing to the same port as @Tratcher suggested

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
```diff
@@ -0,0 +1,134 @@
...
+            // 2. HTTPS_PORT environment variable
+            // 3. IServerAddressesFeature (checked in HttpsRedirectionMiddleware.Invoke
+            // 4. 443 (or not set)
+
+            if (_httpsPort != null)
+            {
+                return;
+            }
+
+            var configHttpsPort = _config["HTTPS_PORT"];
```
Tratcher: `_config.Get<int?>("HTTPS_PORT");`

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
```diff
@@ -0,0 +1,134 @@
...
+                        throw new ArgumentException($"Cannot specify multiple https ports in IServerAddressesFeature. " +
+                            $"Conflict found with ports: {httpsPort.Value}, {serverAddress.Port}.");
+                    }
+                    else
+                    {
+                        httpsPort = serverAddress.Port;
+                    }
+                }
+            }
+            _httpsPort = httpsPort;
```
Tratcher: ?? 443

src/Microsoft.AspNetCore.HttpsPolicy/Microsoft.AspNetCore.HttpsPolicy.csproj
Tratcher: Http.Sys has a similar though simpler parser. There may be a chance to unify all three.
https://github.com/aspnet/HttpSysServer/blob/14aec22cda2c0c81b369b08c7d260afffae20c9d/src/Microsoft.AspNetCore.Server.HttpSys/UrlPrefix.cs

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
Tratcher: Too bad there's no way to populate IServerAddressesFeature for In-Proc ANCM. Right?

src/Microsoft.AspNetCore.HttpsPolicy/Microsoft.AspNetCore.HttpsPolicy.csproj
jkotalik: I would prefer this. Would it be possible to move all Url parsing to HttpAbstractions? cc/ @halter73

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// <param name="context"></param>
+        /// <returns></returns>
+        public Task Invoke(HttpContext context)
+        {
+            if (context.Request.IsHttps)
+            {
+                return _next(context);
+            }
+
+            if (!_evaluatedServerAddressesFeature && !_httpsPort.HasValue)
```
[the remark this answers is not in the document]
javiercn: I would name it Ensure instead of Check if you do what @Tratcher suggests.

src/Microsoft.AspNetCore.HttpsPolicy/Microsoft.AspNetCore.HttpsPolicy.csproj
halter73: Go for it.

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
jkotalik: I'll make an issue for this (to at least investigate whether we can populate it).

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
jkotalik: Api doesn't exist I think: https://github.com/aspnet/Configuration/blob/dc6fbde39f532bf154ee1eb02d8be08a0a1605e4/src/Config.Binder/ConfigurationBinder.cs#L26

jkotalik: GetValue<int?> ?

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
jkotalik: https://github.com/aspnet/IISIntegration/issues/485

jkotalik:
https://github.com/aspnet/HttpAbstractions/pull/977

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionBuilderExtensions.cs
Tratcher: Don't for get @javiercn 's comment about not needing to resolve IConfig here.

jkotalik: whoopz

jkotalik:
This now consumes BindingAddress. 🆙 📅

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
```diff
@@ -0,0 +1,125 @@
...
+        private void CheckForHttpsPorts()
+        {
+            // The IServerAddressesFeature will not be ready until the middleware is Invoked,
+            // Order for finding the HTTPS port:
+            // 1. Set in the HttpsRedirectionOptions
+            // 2. HTTPS_PORT environment variable
+            // 3. IServerAddressesFeature
+            // 4. 443 (or not set)
+
+            var configHttpsPort = _config.GetValue<int?>("HTTPS_PORT");
```
Tratcher: assign directly to _httpsPort, no need for the temp variable.

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
```diff
@@ -0,0 +1,125 @@
...
+            {
+                _httpsPort = configHttpsPort;
+                return;
+            } 
+
+            int? httpsPort = null;
+            foreach (var address in _serverAddressesFeature.Addresses)
+            {
+                var bindingAddress = BindingAddress.Parse(address);
+                if (bindingAddress.Scheme == "https")
```
Tratcher: BindingAddress doesn't normalize case, use string.Equals OrdinalIgnoreCase

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
```diff
@@ -0,0 +1,125 @@
...
+                {
+                    // If we find multiple different https ports specified, throw
+                    if (httpsPort != null && httpsPort != bindingAddress.Port)
```
Tratcher: .HasValue

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs
```diff
@@ -0,0 +1,125 @@
...
+                    {
+                        throw new ArgumentException($"Cannot specify multiple https ports in IServerAddressesFeature. " +
```
Tratcher: "Cannot determine the https port from IServerAddressesFeature, multiple values were found. Please set the desired port explicitly on HttpsRedirectionOptions.HttpsPort."

src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionOptions.cs
```diff
@@ -16,11 +16,8 @@ public class HttpsRedirectionOptions
         public int RedirectStatusCode { get; set; } = StatusCodes.Status302Found;
 
         /// <summary>
-        /// The TLS port to be added to the redirected URL.
+        /// The HTTPS port to be added to the redirected URL.
         /// </summary>
-        /// <remarks>
-        /// Defaults to 443 if not provided.
```
Tratcher: Summarize the fallback algorithm in remarks.

Tratcher:
File followup bugs in Kestrel and HttpSys to consume the new BindingAddress

Issue #279, opened 2017-12-19
Regex Timeout are too short, not consistent, and not configurable in RewriteMiddleware


jkotalik:
Regex timeouts that are used across [UrlRewrite](https://github.com/aspnet/BasicMiddleware/blob/dev/src/Microsoft.AspNetCore.Rewrite/Internal/IISUrlRewrite/UrlRewriteRuleBuilder.cs#L13), ApacheModRewrite, and [code rules](https://github.com/aspnet/BasicMiddleware/blob/03b63e2c2a5ec27deef4be692ff2119188b67d73/src/Microsoft.AspNetCore.Rewrite/Internal/RewriteRule.cs) are not consistent. The are set at 1 second or 1 millisecond in different places. Secondly, 1 millisecond is too short for a regex expression; it should be 1 second as the timeout is mostly for making sure your server isn't locked. Also if the expression timeouts, it returns a 500 server error as the exception is unhandled. Thirdly, the timeout should be configurable in RewriteOptions. 

Plan of action:
- Timeouts should be increased to 1 second across the board.
- For 2.1, we should make Regex timeouts configurable a
- The timeouts should be backported to 2.0 and 1.1

@Eilon @muratg this is a patch candidate. Workarounds are very difficult.

Eilon:
Need to consult with @blowdart on a good timeout. I agree 1ms is waaaay too small.

blowdart:
1.1 milliseconds.

Merry Christmas ⛄️ 

Oh ok, 1 second is fine as a default, as long as it's configurable.

Eilon:
@blowdart but in a patch, we might not make it configurable. Would 1sec be reasonable in the patch? (With a switch to restore the original bizarre 1ms timeout.)

jkotalik:
The patch is a simple fix. However the redesign would require more work. When calling options.AddRedirect(), options.AddIISUrlRewrite, etc, we eagerly create the Regex object, which is too early as options would now store the regex timeout. Someone could do:
```
options.AddRedirect(...); // Creates a regex object with default timeout of 1 ms
options.RegexTimeout = TimeSpan.FromSeconds(1);
options.AddRewrite(...);
```
which would set the timeout of 1 ms instead of 1 second for the regex expression. We shouldn't have options apis that need to be ordered.
We could either create the regex objects on first request or use a different pattern to know when rules are done being added.

blowdart:
[...]

PR #280, opened 2017-12-20
Increases timeout on Regex in Rewrite.

jkotalik:
For #279 . This will most likely be the PR to patch back to 1.1.x and 2.0.x. For 2.1.x or 3.0.x, we may want to redesign/ do something a bit different. 

Note, the regex objects themselves are all private or created in builders themselves, so it is hard/can't be tested without exposing them.

### DIFF
--- a/src/Microsoft.AspNetCore.Rewrite/Internal/ApacheModRewrite/RuleBuilder.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/Internal/ApacheModRewrite/RuleBuilder.cs
@@ -18,7 +18,7 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.ApacheModRewrite
         private UrlMatch _match;
         private CookieActionFactory _cookieActionFactory = new CookieActionFactory();
 
-        private readonly TimeSpan RegexTimeout = TimeSpan.FromMilliseconds(1);
+        private readonly TimeSpan _regexTimeout = TimeSpan.FromSeconds(1);
 
         public ApacheModRewriteRule Build()
         {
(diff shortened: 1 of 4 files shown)


Eilon:
For 2.1 I think this is fine, unless we also want to make it configurable. But for patches I'd want a quirk switch to revert the behavior.

src/Microsoft.AspNetCore.Rewrite/Internal/ApacheModRewrite/RuleBuilder.cs
```diff
@@ -18,7 +18,7 @@ public class RuleBuilder
         private UrlMatch _match;
         private CookieActionFactory _cookieActionFactory = new CookieActionFactory();
 
-        private readonly TimeSpan RegexTimeout = TimeSpan.FromMilliseconds(1);
+        private readonly TimeSpan _regexTimeout = TimeSpan.FromSeconds(1);
```
mikaelm12: 👍

jkotalik:
What do you mean by: "But for patches I'd want a quirk switch to revert the behavior."

Eilon:
If we end up patching this, we would choose one of two options:
1. Change the default to 1s, with a "mode" to revert back to the original 1ms behavior.
2. Leave the behavior as is with a default of 1ms, and have a "mode" to enable the new, longer 1s behavior.

"Quirks" modes are ways to "quirk" the behavior to restore some legacy behavior.

Tratcher:
Let's move this to the 2.0.0 branch so we can see what the real patch would look like. E.g. you wouldn't do these renames in a patch. Also adding the quirks settings.

I expect our 2.1 fix will be substantially different due to adding a new public API.

jkotalik:
I'm going to merge this for now and then make a PR to the 2.0.x and 1.1.x branches with the quirk mode. We can address the refactor for the 2.1.x branch at a different point.

Issue #210, opened 2017-02-16
Handle redirect http consistently with MVC RequireHttpsAttribute


javiercn:
Per discussion with @Tratcher, the url rewrite middleware supports rewriting urls for functionality similar to require https. We should handle the scenario in a consistent way no matter what option (url rewrite or require https) we decide to use.

Tratcher:
Specifically do not redirect POST or similar requests with bodies. MVC sends a 403 Forbidden response for these requests.

glennc:
@Tratcher Can I get more details on what the differences between these two are?

Tratcher:
Redirecting for POST requests was the big inconstancy we noticed. MVC won't redirect a POST request, it returns 403 instead. Rewrite does redirect, and this can cause clients to convert the POST to a GET.

muratg:
From the triage room, @Tratcher thinks using status code 306 or 307 could be another way to fix this issue.

muratg:
@Tratcher will investigate some simple things we can do in 2.1.  This would technically be a breaking change, but it may be something that we can accept because of the already broken scenarios. 

If simple fixes are not applicable, we'll backlog.

javiercn:
@muratg @Tratcher Do we really need to do anything here? Given our plans for HTTPS in 2.1 I would avoid doing anything here. You either/or the new HTTPS stuff or the RedirectToHttps functionality. I don't think we need to support anything here, specially at the cost of a breaking change. I would like to see RedirectToHttpsAttribute vanish from existence in the future as much as posible.

Tratcher:
@javiercn the main question is how correct is the implementation in this repo. RedirectToHttpsAttribute is merely an interesting reference. 301/302 is a problem for POST requests.

javiercn:
@Tratcher I understand 301/302 is a problem for POST requests, but who in real life sends 307/308 as a redirect. I'm ok if the redirect only works for get requests. I think it's better to 400 post requests as you don't want that to silently work (given that the information has already been posted through an insecure channel).

Tratcher:
I don't know yet, that's why this marked as Investigate.

Tratcher:
307 does look like the correct default. I've verified it works with Chrome, IE, Edge, HttpClient Core (Win) and .NET, and Curl (Win). I'll send a PR.

javiercn:
@Tratcher Are you planning to 307 by default on POST requests?

Tratcher:
On all requests.

PR #285, opened 2018-01-03
Https Redirect: Change default status code to 307

Tratcher:
#210 
307 indicates that the method should not be changed when redirecting as was common practice for 301. This works better for request with bodies.


jkotalik:
Should we also make this the default in UrlRewrite? https://github.com/aspnet/BasicMiddleware/blob/9814f3bfbcf3a11dc86c4047fbfc2fe48744f96c/src/Microsoft.AspNetCore.Rewrite/RewriteOptionsExtensions.cs#L96

Tratcher:
I hesitate to change the default on previously shipped components. There is some small risk that this won't work with some clients.

jkotalik:
Alright, that's fine. Can you add a remark saying the default is 307?

Issue #292, opened 2018-01-30
Support REQUEST_METHOD variable in IIS URL Rewrite file parsing


mikaelm12:
See: https://github.com/aspnet/BasicMiddleware/issues/191#issuecomment-361410542

mikaelm12:
Done: b85d69c882fbd6aa2d6b0caa4b5b59d58f407140

PR #294, opened 2018-01-30
Add support for REQUEST_METHOD server variable

mikaelm12:
For issue: https://github.com/aspnet/BasicMiddleware/issues/292

cc @Tratcher @muratg


muratg:
I think this would be good to get in into preview1, but could you start an internal thread with @DamianEdwards and @Eilon for official approval?

Issue #194, opened 2016-12-21
Add method to RewriteOptions to modify www to non-www and vice-versa.


jkotalik:
Seems like a use case that should be inherently supported. Though most rewriters do not support regex match on host name, if we can support it with a method, that'd be great!

muratg:
Sounds like a good feature!  We'll assess this in the future based on the customer feedback.

muratg:
http://stackoverflow.com/questions/41205435/asp-net-core-1-1-url-rewriting-www-to-non-www

GuardRex:
An option for matching on the entire URL would be nice here, including scheme and even fragments.

HarelM:
See also #220, rewrite entire URL including parameters.
This is essential in order to be able to serve facebook crawler the right details when using SPA.
I have copy-paste the code from the rewrite rule on my site but I don't think it's the right approach.
the following is the rule I would like to remove from my web.config:
Current code can't do it.
```xml
<rewrite>		    
   <rules>		      
	 <rule name="opengraph" patternSyntax="ECMAScript" stopProcessing="true">	
	   <match url=".*" />		          
	   <action type="Rewrite" url="/api/opengraph/{C:1}" logRewrittenUrl="true" />		      
	   <conditions logicalGrouping="MatchAny" trackAllCaptures="true">		         
		 <add input="{QUERY_STRING}" pattern=".*escaped_fragment_=%2F%3Fs%3D(.*)" />		          
	   </conditions>		         
	 </rule>		     
   </rules>		  
</rewrite>
```

HarelM:
Any updates on this? Is there a plan to support this request?

jkotalik:
@muratg Could this be something to add in 2.1?

HarelM:
Any updates on this?

alvipeo:
Yeah, it's been a while...

HarelM:
Pinging... I still have a piece of code I would like to get rid of and am waiting for this to be solved so remove it.

muratg:
@glennc

muratg:
@glennc Do you think this is something we'd consider in 2.1?

@HarelM @alvipeo -- would one of you guys be interested in sending a PR?

HarelM:
sure, any guidelines? Tests locating? Preferred implementation method?

jkotalik:
[...]

PR #297, opened 2018-02-07
Adding RedirectToWwwRule and extension

clockworkmice:
Adding a simple way to enforce www. subdomain using a new redirect rule

Addresses #194

### DIFF
--- a/src/Microsoft.AspNetCore.Rewrite/Extensions/RewriteMiddlewareLoggingExtensions.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/Extensions/RewriteMiddlewareLoggingExtensions.cs
@@ -87,7 +87,7 @@ namespace Microsoft.AspNetCore.Rewrite.Logging
             _redirectedToWww = LoggerMessage.Define(
                             LogLevel.Information,
                             13,
-                            "Request redirected to WWW");
+                            "Request redirected to www");
         }
 
         public static void RewriteMiddlewareRequestContinueResults(this ILogger logger, string currentUrl)
--- a/src/Microsoft.AspNetCore.Rewrite/Extensions/RewriteMiddlewareLoggingExtensions.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/Extensions/RewriteMiddlewareLoggingExtensions.cs
@@ -16,6 +16,7 @@ namespace Microsoft.AspNetCore.Rewrite.Logging
         private static readonly Action<ILogger, Exception> _modRewriteDidNotMatchRule;
         private static readonly Action<ILogger, Exception> _modRewriteMatchedRule;
         private static readonly Action<ILogger, Exception> _redirectedToHttps;
+        private static readonly Action<ILogger, Exception> _redirectedToWww;
         private static readonly Action<ILogger, string, Exception> _redirectSummary;
         private static readonly Action<ILogger, string, Exception> _rewriteSummary;
         private static readonly Action<ILogger, string, Exception> _abortedRequest;
@@ -82,6 +83,11 @@ namespace Microsoft.AspNetCore.Rewrite.Logging
                             LogLevel.Debug,
                             12,
                             "Request to {requestedUrl} was ended");
+
+            _redirectedToWww = LoggerMessage.Define(
+                            LogLevel.Information,
+                            13,
+                            "Request redirected to WWW");
         }
 
         public static void RewriteMiddlewareRequestContinueResults(this ILogger logger, string currentUrl)
@@ -122,6 +128,11 @@ namespace Microsoft.AspNetCore.Rewrite.Logging
         public static void RedirectedToHttps(this ILogger logger)
         {
             _redirectedToHttps(logger, null);
+        }
+
+        public static void RedirectedToWww(this ILogger logger)
+        {
+            _redirectedToWww(logger, null);
         }
 
         public static void RedirectedSummary(this ILogger logger, string redirectedUrl)
--- a/src/Microsoft.AspNetCore.Rewrite/Internal/RedirectToWwwRule.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/Internal/RedirectToWwwRule.cs
@@ -1,0 +1,39 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Text;
+using Microsoft.AspNetCore.Rewrite.Logging;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNetCore.Rewrite.Internal
+{
+    public class RedirectToWwwRule : IRule
+    {
+        public int StatusCode { get; set; }
+        public bool AllowLocalhost { get; set; } = true;
+
+        public virtual void ApplyRule(RewriteContext context)
+        {
+            var req = context.HttpContext.Request;
+            if (AllowLocalhost && req.Host.Host.StartsWith("localhost", StringComparison.OrdinalIgnoreCase))
+            {
+                context.Result = RuleResult.ContinueRules;
+                return;
+            }
+
+            if (req.Host.Value.StartsWith("www.", StringComparison.OrdinalIgnoreCase))
+            {
+                context.Result = RuleResult.ContinueRules;
+                return;
+            }
+
+            var newUrl = new StringBuilder().Append(req.Scheme).Append("://www.").Append(req.Host).Append(req.PathBase).Append(req.Path).Append(req.QueryString);
+            var response = context.HttpContext.Response;
+            response.StatusCode = StatusCode;
+            response.Headers[HeaderNames.Location] = newUrl.ToString();
+            context.Result = RuleResult.EndResponse;
+            context.Logger?.RedirectedToWww();
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Rewrite/Internal/RedirectToWwwRule.cs
+++ b/src/Microsoft.AspNetCore.Rewrite/Internal/RedirectToWwwRule.cs
@@ -2,7 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Text;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.AspNetCore.Rewrite.Logging;
 using Microsoft.Net.Http.Headers;
 
@@ -10,13 +11,17 @@ namespace Microsoft.AspNetCore.Rewrite.Internal
 {
     public class RedirectToWwwRule : IRule
     {
-        public int StatusCode { get; set; }
-        public bool AllowLocalhost { get; set; } = true;
+        public readonly int _statusCode;
+
+        public RedirectToWwwRule(int statusCode)
+        {
+            _statusCode = statusCode;
+        }
 
         public virtual void ApplyRule(RewriteContext context)
         {
             var req = context.HttpContext.Request;
-            if (AllowLocalhost && req.Host.Host.StartsWith("localhost", StringComparison.OrdinalIgnoreCase))
+            if (req.Host.Host.Equals("localhost", StringComparison.OrdinalIgnoreCase))
             {
                 context.Result = RuleResult.ContinueRules;
                 return;
@@ -28,10 +33,11 @@ namespace Microsoft.AspNetCore.Rewrite.Internal
                 return;
             }
 
-            var newUrl = new StringBuilder().Append(req.Scheme).Append("://www.").Append(req.Host).Append(req.PathBase).Append(req.Path).Append(req.QueryString);
+            var wwwHost = new HostString($"www.{req.Host.Value}");
+            var newUrl = UriHelper.BuildAbsolute(req.Scheme, wwwHost, req.PathBase, req.Path, req.QueryString);
             var response = context.HttpContext.Response;
-            response.StatusCode = StatusCode;
-            response.Headers[HeaderNames.Location] = newUrl.ToString();
+            response.StatusCode = _statusCode;
+            response.Headers[HeaderNames.Location] = newUrl;
             context.Result = RuleResult.EndResponse;
             context.Logger?.RedirectedToWww();
         }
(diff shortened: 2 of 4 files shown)


src/Microsoft.AspNetCore.Rewrite/Internal/RedirectToWwwRule.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    public class RedirectToWwwRule : IRule
+    {
+        public int StatusCode { get; set; }
+        public bool AllowLocalhost { get; set; } = true;
+
+        public virtual void ApplyRule(RewriteContext context)
+        {
+            var req = context.HttpContext.Request;
+            if (AllowLocalhost && req.Host.Host.StartsWith("localhost", StringComparison.OrdinalIgnoreCase))
```
jkotalik: I don't see why you need to do StartsWith here.

src/Microsoft.AspNetCore.Rewrite/Internal/RedirectToWwwRule.cs
```diff
@@ -0,0 +1,39 @@
...
+                return;
+            }
+
+            if (req.Host.Value.StartsWith("www.", StringComparison.OrdinalIgnoreCase))
+            {
+                context.Result = RuleResult.ContinueRules;
+                return;
+            }
+
+            var newUrl = new StringBuilder().Append(req.Scheme).Append("://www.").Append(req.Host).Append(req.PathBase).Append(req.Path).Append(req.QueryString);
```
jkotalik: Use UriHelpers.BuildAbsolute instead here. https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http.Extensions/UriHelper.cs#L48

src/Microsoft.AspNetCore.Rewrite/RewriteOptionsExtensions.cs
```diff
@@ -117,5 +117,36 @@ public static RewriteOptions AddRedirectToHttps(this RewriteOptions options, int
...
+            return AddRedirectToWww(options, statusCode: StatusCodes.Status301MovedPermanently);
+        }
+
+        /// <summary>
+        /// Redirect a request to www subdomain if the incoming request is non-www
+        /// </summary>
+        /// <param name="options">The <see cref="RewriteOptions"/>.</param>
+        public static RewriteOptions AddRedirectToWww(this RewriteOptions options)
+        {
+            return AddRedirectToWww(options, statusCode: StatusCodes.Status302Found);
```
jkotalik: I think we are using a redirect status code of 307 from now on. @Tratcher ?

src/Microsoft.AspNetCore.Rewrite/Internal/RedirectToWwwRule.cs
clockworkmice: As opposed to straight equals? I think you're right actually, I was using Value previously

src/Microsoft.AspNetCore.Rewrite/Internal/RedirectToWwwRule.cs
```diff
@@ -0,0 +1,42 @@
...
+            }
+
+            if (req.Host.Value.StartsWith("www.", StringComparison.OrdinalIgnoreCase))
+            {
+                context.Result = RuleResult.ContinueRules;
+                return;
+            }
+
+            var wwwHost = new HostString($"www.{req.Host}");
+            var newUrl = UriHelper.BuildAbsolute(req.Scheme, wwwHost, req.PathBase, req.Path, req.QueryString);
```
jkotalik: Add in the FragmentString here (req.FragmentString). It was a mistake for other rules to not include it.

src/Microsoft.AspNetCore.Rewrite/RewriteOptionsExtensions.cs
```diff
@@ -117,5 +117,36 @@ public static RewriteOptions AddRedirectToHttps(this RewriteOptions options, int
             options.Rules.Add(new RedirectToHttpsRule { StatusCode = statusCode, SSLPort = sslPort });
             return options;
         }
+
+        /// <summary>
+        /// Redirect a request to www subdomain if the incoming request is non-www, with returning a 301
+        /// status code for permanently redirected.
+        /// </summary>
```
jkotalik: I'd reword to 
``` 
Permanently redirects the request to the www subdomain if the request is is non-www.
```

src/Microsoft.AspNetCore.Rewrite/RewriteOptionsExtensions.cs
```diff
@@ -117,5 +117,36 @@ public static RewriteOptions AddRedirectToHttps(this RewriteOptions options, int
...
+        /// <param name="options">The <see cref="RewriteOptions"/>.</param>
+        /// <returns></returns>
+        public static RewriteOptions AddRedirectToWwwPermanent(this RewriteOptions options)
+        {
+            return AddRedirectToWww(options, statusCode: StatusCodes.Status301MovedPermanently);
+        }
+
+        /// <summary>
+        /// Redirect a request to www subdomain if the incoming request is non-www
```
jkotalik:
```
Redirect the request to the www subdomain if the incoming request is non-www.
```

src/Microsoft.AspNetCore.Rewrite/RewriteOptionsExtensions.cs
```diff
@@ -117,5 +117,36 @@ public static RewriteOptions AddRedirectToHttps(this RewriteOptions options, int
...
+        /// </summary>
+        /// <param name="options">The <see cref="RewriteOptions"/>.</param>
+        /// <param name="statusCode">The status code to add to the response.</param>
+        public static RewriteOptions AddRedirectToWww(this RewriteOptions options, int statusCode)
+        {
+            options.Rules.Add(new RedirectToWwwRule { StatusCode = statusCode });
```
jkotalik: Is there a way to set AllowLocalhost? I think you need to add another extension:
```csharp
 public static RewriteOptions AddRedirectToWww(this RewriteOptions options, int statusCode, bool allowLocalhost)
```

src/Microsoft.AspNetCore.Rewrite/Internal/RedirectToWwwRule.cs
```diff
@@ -0,0 +1,42 @@
...
+using Microsoft.AspNetCore.Http.Extensions;
+using Microsoft.AspNetCore.Rewrite.Logging;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNetCore.Rewrite.Internal
+{
+    public class RedirectToWwwRule : IRule
+    {
+        public int StatusCode { get; set; }
+        public bool AllowLocalhost { get; set; } = true;
```
jkotalik: Bleh, I don't like how I designed this when I was an intern. Let's make StatusCode and AllowLocalhost private and set them in the constructor, like:
```csharp
public RedirectToWwwRule(int statusCode, bool allowLocalhost) ...
```

test/Microsoft.AspNetCore.Rewrite.Tests/MiddlewareTests.cs
```diff
@@ -147,6 +147,93 @@ public class MiddlewareTests
...
+            {
+                app.UseRewriter(options);
+            });
+            var server = new TestServer(builder);
+
+            var response = await server.CreateClient().GetAsync(new Uri("https://localhost:443"));
+
+            Assert.Null(response.Headers.Location);
+        }
+
```
jkotalik: Let's test a few more cases:
- Uri with http
- Uri with many parts, like "https://example.com/test?a=b#foobar" and verify www is added appropriately.
- Uri with just localhost (without port).

You can do this easily by making the test a Theory. See https://github.com/aspnet/BasicMiddleware/blob/dev/test/Microsoft.AspNetCore.HttpsPolicy.Tests/HstsMiddlewareTests.cs#L57

src/Microsoft.AspNetCore.Rewrite/Internal/RedirectToWwwRule.cs
clockworkmice: That's I would have written it personally too. I was just trying to be consistent with RedirectToHttpsRule. Will change

src/Microsoft.AspNetCore.Rewrite/Internal/RedirectToWwwRule.cs
clockworkmice: how would I get the # fragment at this point? We're on the server

src/Microsoft.AspNetCore.Rewrite/RewriteOptionsExtensions.cs
clockworkmice: If someone really wants to not make an exception for localhost, it can be set by adding the RedirectToWwwRule explicitly. I imagined it would be used so infrequently that I didn't include it in an extension method.

Having said that, a comment or even an optional param for the extension method would show the expected behaviour more though when someone is using it for the first time and wondering why it isn't working in IISExpress. What do you think?

src/Microsoft.AspNetCore.Rewrite/Internal/RedirectToWwwRule.cs
```diff
@@ -0,0 +1,42 @@
...
+                return;
+            }
+
+            if (req.Host.Value.StartsWith("www.", StringComparison.OrdinalIgnoreCase))
+            {
+                context.Result = RuleResult.ContinueRules;
+                return;
+            }
+
+            var wwwHost = new HostString($"www.{req.Host}");
```
Tratcher: Use Host.Value here. The encoding is different from Host.ToString.

src/Microsoft.AspNetCore.Rewrite/Internal/RedirectToWwwRule.cs
Tratcher: No, there's never a fragment on the request.

src/Microsoft.AspNetCore.Rewrite/RewriteOptionsExtensions.cs
```diff
@@ -117,5 +117,36 @@ public static RewriteOptions AddRedirectToHttps(this RewriteOptions options, int
...
+
+        /// <summary>
+        /// Redirect a request to www subdomain if the incoming request is non-www, with returning a 301
+        /// status code for permanently redirected.
+        /// </summary>
+        /// <param name="options">The <see cref="RewriteOptions"/>.</param>
+        /// <returns></returns>
+        public static RewriteOptions AddRedirectToWwwPermanent(this RewriteOptions options)
+        {
+            return AddRedirectToWww(options, statusCode: StatusCodes.Status301MovedPermanently);
```
Tratcher: Let's use 308 for new code. It works better for POSTs.

src/Microsoft.AspNetCore.Rewrite/RewriteOptionsExtensions.cs
Tratcher: An non-exposed option might as well not exist. Add an overload or remove the option.

clockworkmice: An optional param on all the extension methods or just this one?

Tratcher: An additional overload with all options. Optional params are discouraged in our public API.

clockworkmice: That's fair enough. Just seems like a lot of bloat for something that won't ever be set to false. Maybe keep the localhost logic but just not make it an option? What do you think? I'm happy with either

Tratcher: Sure, go ahead and remove the option.

clockworkmice:
@Tratcher @jkotalik updated as per comments

src/Microsoft.AspNetCore.Rewrite/Extensions/RewriteMiddlewareLoggingExtensions.cs
```diff
@@ -82,6 +83,11 @@ static RewriteMiddlewareLoggingExtensions()
                             LogLevel.Debug,
                             12,
                             "Request to {requestedUrl} was ended");
+
+            _redirectedToWww = LoggerMessage.Define(
+                            LogLevel.Information,
+                            13,
+                            "Request redirected to WWW");
```
Tratcher: lower case www

src/Microsoft.AspNetCore.Rewrite/Internal/RedirectToWwwRule.cs
```diff
@@ -0,0 +1,45 @@
...
+                return;
+            }
+
+            if (req.Host.Value.StartsWith("www.", StringComparison.OrdinalIgnoreCase))
+            {
+                context.Result = RuleResult.ContinueRules;
+                return;
+            }
+
+            var wwwHost = new HostString($"www.{req.Host.Value}");
```
Tratcher: You'll get some pretty strange results if the host is 127.0.0.1:80 or [::1]:80. I don't think those are common enough to worry about though.

src/Microsoft.AspNetCore.Rewrite/Extensions/RewriteMiddlewareLoggingExtensions.cs
clockworkmice: done

src/Microsoft.AspNetCore.Rewrite/Internal/RedirectToWwwRule.cs
jkotalik: We may want to consider adding a list of excluded hosts, like https://github.com/aspnet/BasicMiddleware/commit/892cebb27eb1487a10e07d70967e490ecb38bce8#diff-15f19c319158f42c2602a43b53630d87R43.

jkotalik:
@clockworkmice thanks again for the contribution. This should be available by 2.1.0-preview2 or on our nightly builds shortly.

src/Microsoft.AspNetCore.Rewrite/Internal/RedirectToWwwRule.cs
Tratcher: No, if it came up you'd want to exclude IP addresses as a class of host, not exclude individual values.

PR #312, opened 2018-03-14
New middleware for host filtering

Tratcher:
This takes the sample we provided for the 2.0 patch and makes it an official package. It also adds a similar host filtering feature to HttpOverrides for x-forwarded-host.

The only outstanding design point is providing a way to switch back and forth between a local server and a reverse proxy configuration. I'll followup on that as a separate work item as it requires coordinating across several components.


src/Microsoft.AspNetCore.HostFiltering/HostFilteringMiddleware.cs
```diff
@@ -0,0 +1,175 @@
...
+
+namespace Microsoft.AspNetCore.HostFiltering
+{
+    /// <summary>
+    /// A middleware used to filter requests by their Host header.
+    /// </summary>
+    public class HostFilteringMiddleware
+    {
+        // Matches Http.Sys.
+        private static readonly byte[] DefaultResponse = Encoding.ASCII.GetBytes(
```
davidfowl: Ugh, since when do we hard code error pages into our middleware? Shouldn't that be handled separately? What if the client is an API client? Why isn't the 400 enough?

Tratcher: There are comments about this elsewhere in the code. This is more severe than your usual failure, it's a binding/protocol level error that has security implications. If you don't provide a body then components like the StatusCode middleware can still generate insecure content. This is one step short of calling Abort.

As for WebAPIs, for an error this severe being able to consume the body is our last concern. That said, I did forget to set the content-type.

Tratcher: @shirhatti

src/Microsoft.AspNetCore.HostFiltering/Microsoft.AspNetCore.HostFiltering.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>
+      ASP.NET Core middleware for filtering out requests with unknown http Host headers.
```
Eilon: `HTTP host`?

src/Microsoft.AspNetCore.HttpOverrides/ForwardedHeadersOptions.cs
```diff
@@ -67,6 +67,20 @@ public class ForwardedHeadersOptions
         /// </summary>
         public IList<IPNetwork> KnownNetworks { get; } = new List<IPNetwork>() { new IPNetwork(IPAddress.Loopback, 8) };
 
+        /// <summary>
+        /// The allowed values from x-forwarded-host. If the list is empty then all hosts are allowed.
+        /// Failing to restrict this these values may allow an attacker to spoof links generated by your service. 
+        /// </summary>
+        /// <remarks>
+        /// - Port numbers must be excluded
```
Eilon: I don't think this will be formatted correctly in doc comments. There's some kind of `<list>` element you need to use. (Kind of like HTML, but different.)

test/Microsoft.AspNetCore.HostFiltering.Tests/HostFilteringMiddlewareTests.cs
```diff
@@ -0,0 +1,310 @@
+﻿using System;
+using System.Threading.Tasks;
```
Eilon: Copyright/license

src/Microsoft.AspNetCore.HostFiltering/HostFilteringMiddleware.cs
```diff
@@ -0,0 +1,175 @@
...
+            }
+
+            _logger.LogDebug("Allowed hosts: " + string.Join("; ", allowedHosts));
+            _allowedHosts = allowedHosts;
+        }
+
+        // returns false if any wildcards were found
+        private bool TryProcessHosts(IEnumerable<string> incoming, IList<StringSegment> results)
+        {
+            foreach (var entry in incoming)
```
jkotalik: Nit: This method is doing a decent amount. I would separate this method into two methods: one that checks if any of the strings are wildcards, and another that transforms the entries to HostString.ToUriComponent.

src/Microsoft.AspNetCore.HostFiltering/HostFilteringOptions.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// Http/1.1 requires a host header, but says the value may be empty.
+        /// </remarks>
+        public bool AllowEmptyHosts { get; set; } = true;
+
+        // Note if this were disabled then things like the status code middleware may try to re-execute
+        // the request. This is a low level protocol violation, pretty error pages should not be required.
+        /// <summary>
+        /// Indicates if the 400 response should include a default message or be empty. This is enabled by default.
+        /// </summary>
+        public bool IncludedFailureMessage { get; set; } = true;
```
jkotalik: Nit: IncludeFailureMessage?

src/Microsoft.AspNetCore.HostFiltering/HostFilteringOptions.cs
```diff
@@ -0,0 +1,42 @@
...
+        public IList<string> AllowedHosts { get; set; } = new List<string>();
+
+        /// <summary>
+        /// Indicates if requests without hosts are allowed. The default is true.
+        /// </summary>
+        /// <remarks>
+        /// HTTP/1.0 does not require a host header.
+        /// Http/1.1 requires a host header, but says the value may be empty.
+        /// </remarks>
+        public bool AllowEmptyHosts { get; set; } = true;
```
jkotalik: Why do we prefer to have the default of true here? Because it isn't required in 1.0?

src/Microsoft.AspNetCore.HttpOverrides/ForwardedHeadersMiddleware.cs
```diff
@@ -95,6 +102,37 @@ private static void EnsureOptionNotNullorWhitespace(string value, string propert
             }
         }
 
+        private void PreProcessHosts()
+        {
+            if (_options.AllowedHosts == null || _options.AllowedHosts.Count == 0)
```
jkotalik: Nit: Share between the middleware somehow? Add to HttpAbstractions?

test/Microsoft.AspNetCore.HostFiltering.Tests/HostFilteringMiddlewareTests.cs
```diff
@@ -0,0 +1,310 @@
...
+            var response = await server.CreateClient().GetAsync("/");
+            Assert.Equal(status, (int)response.StatusCode);
+        }
+
+        [Theory]
+        [InlineData("localHost", "localhost")]
+        [InlineData("localHost", "*")] // Any - Used by HttpSys
+        [InlineData("localHost", "[::]")] // IPv6 Any - This is what Kestrel reports when binding to *
+        [InlineData("localHost", "0.0.0.0")] // IPv4 Any
+        [InlineData("localhost:9090", "example.com;localHost")]
```
jkotalik: Where is the example.com coming from here?

test/Microsoft.AspNetCore.HostFiltering.Tests/HostFilteringMiddlewareTests.cs
```diff
@@ -0,0 +1,310 @@
...
+                });
+            var server = new TestServer(builder);
+            var response = await server.CreateRequest("/").GetAsync();
+            Assert.Equal(200, (int)response.StatusCode);
+        }
+
+        [Theory]
+        [InlineData("example.com", "localhost")]
+        [InlineData("localhost:9090", "example.com;")]
+        [InlineData(";", "example.com;localhost")]
```
jkotalik: Split on semicolons in your test as you do it here.

samples/HostFilteringSample/Startup.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            Config = config;
+        }
+
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddHostFiltering(options =>
+            {
+                // If this is excluded then it will fall back to the server's addresses
+                options.AllowedHosts = Config["AllowedHosts"]?.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
```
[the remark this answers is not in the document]
Tratcher: We make a point of not hardcoding config in middleware. The Https redirect was the rare exception.

src/Microsoft.AspNetCore.HostFiltering/HostFilteringMiddleware.cs
shirhatti: Gotta agree with Chris here. Best to do what http.sys does

samples/HostFilteringSample/Startup.cs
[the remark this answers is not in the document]
Tratcher: I found a cleaner API

test/Microsoft.AspNetCore.HostFiltering.Tests/HostFilteringMiddlewareTests.cs
Tratcher: example.com is a separate allowed host. I wanted to make sure it would test the whole list.

src/Microsoft.AspNetCore.HttpOverrides/ForwardedHeadersMiddleware.cs
Tratcher: Not planning on this right now, there are some minor differences.

Tratcher:
Updated

samples/HostFilteringSample/Startup.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            Config = config;
+        }
+
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddHostFiltering(options =>
+            {
+                // If this is excluded then it will fall back to the server's addresses
+                options.AllowedHosts = Config.GetSection("AllowedHosts").Get<List<string>>();
```
jkotalik: This is much better.

src/Microsoft.AspNetCore.HostFiltering/HostFilteringMiddleware.cs
```diff
@@ -0,0 +1,182 @@
...
+        {
+            return (string.Equals("*", host, StringComparison.Ordinal) // HttpSys wildcard
+                           || string.Equals("[::]", host, StringComparison.Ordinal) // Kestrel wildcard, IPv6 Any
+                           || string.Equals("0.0.0.0", host, StringComparison.Ordinal)); // IPv4 Any
+        }
+
+        // This does not duplicate format validations that are expected to be performed by the host.
+        private bool CheckHost(HttpContext context)
+        {
+            var host = new StringSegment(context.Request.Headers[HeaderNames.Host].ToString()).Trim();
```
jkotalik: As a general question for host validation, do we want to check the request line for the host?

src/Microsoft.AspNetCore.HostFiltering/HostFilteringMiddleware.cs
```diff
@@ -0,0 +1,182 @@
...
+                }
+                if (_logger.IsEnabled(LogLevel.Debug))
+                {
+                    _logger.LogDebug($"{context.Request.Protocol} request allowed with missing or empty host header.");
+                }
+                return true;
+            }
+
+            if (_allowAnyNonEmptyHost == true)
+            {
```
jkotalik: Nit add a log here to say that the host header was accepted.

src/Microsoft.AspNetCore.HostFiltering/HostFilteringMiddleware.cs
Tratcher: We don't have access to that, only the server sees it.

src/Microsoft.AspNetCore.HostFiltering/HostFilteringMiddleware.cs
```diff
@@ -0,0 +1,182 @@
...
+                if (!TryProcessHosts(_options.AllowedHosts, allowedHosts))
+                {
+                    _logger.LogDebug("Wildcard detected, all requests with hosts will be allowed.");
+                    _allowAnyNonEmptyHost = true;
+                    return;
+                }
+            }
+            else
+            {
+                if (!TryProcessHosts(_serverAddresses.Addresses.Select(address => BindingAddress.Parse(address).Host), allowedHosts))
```
Tratcher: @shirhatti @davidfowl thinking about this more I realized that the use of IServerAddresses here is nonsensical. You need the middleware when there's something deficient about the server configuration, so the middleware will always need different values than the server. E.g. for the primary use case Kestrel, even if you specify https://foo.com:443/ it actually populates IServerAddresses with https://[::]:443/ which is useless to us. Copying config from ANCM in proc or HttpSys is also useless because you'd only be repeating the same validation they'd already done (or none if they were set to wildcards). For the in-proc scenarios you need correctly configure the server and exclude this middleware or set it to * to disable it.

I'm going to remove IServerAddresses from this middleware.

natemcmaster:
@Tratcher CI automation is failing to take in this change to Universe. Can you follow through with the rest of these steps? https://github.com/aspnet/coherence-signed/wiki/New-package-Id-creation

Tratcher:
@pranavkm this looks like a band merge. The changes were checked into dev, branched to release/2.1, but somehow dropped when release was merged back to dev.
https://github.com/aspnet/Universe/commit/fa825fe24747dc311c396e092a656fd3a086220d
https://github.com/aspnet/Universe/commits/release/2.1/build/artifacts.props
https://github.com/aspnet/Universe/commits/dev/build/artifacts.props

I'll cherry pick it back to dev.

PR #313, opened 2018-03-14
Create or update baselines to match 2.0

ryanbrandenburg:


src/Microsoft.AspNetCore.Buffering/baseline.netcore.json
```diff
@@ -0,0 +1,66 @@
+{
+  "AssemblyIdentity": "Microsoft.AspNetCore.Buffering, Version=2.0.2.0, Culture=neutral, PublicKeyToken=adb9793829ddae60",
```
Tratcher: ?? This is a prototype, there is no version 2.0.2.

Eilon: Hmm yes that's an odd version #...

Eilon: As in, it should be 0.4.0 or similar...

Tratcher: Can we remove baselines for prototypes?

Eilon: @Tratcher yeah good point I think that is the right thing to do. There can't be a breaking change in a pre-1.0.0 package.

ryanbrandenburg:
:up::date:

Issue #121, opened 2016-10-06
Support mime-type wildcards


Tratcher:
In IIS a dev can specify wildcard mime types when checking if it should compress the response. Right now the middleware can only do exact matches. It would be much simpler if devs could specify wildcards.

Note in IIS it can opt in or out with wildcards, and override with specific values. E.g. text/\* can be enabled and text/foo can be disabled. Most specific match wins.

RE:
https://www.iis.net/configreference/system.webserver/httpcompression/dynamictypes
IIS/Express Defaults:

```
<httpCompression
      directory="%SystemDrive%\inetpub\temp\IIS Temporary Compressed Files">
   <scheme name="gzip" dll="%Windir%\system32\inetsrv\gzip.dll" />
   <dynamicTypes>
      <add mimeType="text/*" enabled="true" />
      <add mimeType="message/*" enabled="true" />
      <add mimeType="application/javascript" enabled="true" />
      <add mimeType="*/*" enabled="false" />
   </dynamicTypes>
   <staticTypes>
      <add mimeType="text/*" enabled="true" />
      <add mimeType="message/*" enabled="true" />
      <add mimeType="application/javascript" enabled="true" />
      <add mimeType="*/*" enabled="false" />
   </staticTypes>
</httpCompression>
```

Yves57:
When I wrote the initial PR of the Response Compression middleware, I thought about wildcards and other MIME type filtering problems.

If I can do a suggestion, I propose a class like that:

```
public class MimeTypeFilter
{
   public void Add(string mimeType, bool enable) { ... }
   public void Remove(string mimeType) { ... }
   public void Clear() { ... }
   public bool IsMatch(string mimeType) { ... }
}
```

- The class may be useful in other modules, so it may be added in `Microsoft.AspNetCore.WebUtilities`.
- I call the check method `IsMatch` (like in `Regex`) and not `Contains` because the wildcards make `MimeTypeFilter` not strictly a 'simple' container.
- To avoid conflicts between Enable and Disable rules (for instance `f.Add("text/*", true); f.Add("text/plain", false);`), I see two solutions:
   * Keep the `Add()` calls order, and so it checks the rules from first to last. But I think it would not be optimal in terms of performance.
   * My prefered solution: Apply the Enable rules first, and then apply the Disable rules.
- * (star) is accepted only for the subtype part. Exception: `Add("*/*", true);`
[...]

PR #314, opened 2018-03-17
MimeTypeFilter instead of allowed MimeType list

andrewslavin:
Provide more granular control over the MIME types of HTTP responses that should or should not be compressed, via populating `ResponseCompressionOptions.MimeTypeFilter`, as described here: #121.


andrewslavin:
@Tratcher @glennc Please have a look.

src/Microsoft.AspNetCore.ResponseCompression/MimeTypeFilter.cs
```diff
@@ -0,0 +1,207 @@
...
+            set.Clear();
+            HasBeenPopulated = true;
+            return this;
+        }
+
+        //This is NOT a proper MIME type validation. It only checks conformance to the format
+        //"type/sybtype", "type/*" or "*/*", for the purpose of being used in the lists of MIME
+        //types that should and should not be compressed.
+        private static bool IsValid(string mimeType) =>
+            Regex.IsMatch(mimeType, @"^([^\s\*/]+/([^\s\*/]+|\*)|\*/\*)$", RegexOptions.Compiled);
```
Tratcher: Don't use regex for validation. We could use MediaTypeHeaderValue.TryParse, or just not worry about the format, we're only matching against it, not sending it to the client.

src/Microsoft.AspNetCore.ResponseCompression/MimeTypeFilter.cs
```diff
@@ -0,0 +1,207 @@
...
+
+        private bool? ShouldCompressExact(string mimeType)
+        {
+            //check not compressed first, then compressed
+            if (_notCompressed.Contains(mimeType))
```
Tratcher: With this approach a notCompressed wildcard (image/*) would loose vrs an exact _compressed entry (image/foo). That makes wildcards in notCompressed ineffective and you might as well not check for it.

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionOptions.cs
```diff
@@ -13,8 +14,14 @@ public class ResponseCompressionOptions
         /// <summary>
         /// Response Content-Type MIME types to compress.
         /// </summary>
+        [Obsolete("Use MimeTypeFilter.")]
```
Tratcher: Any way to map this through? I don't think we can obsolete this in 2.1.

src/Microsoft.AspNetCore.ResponseCompression/MimeTypeFilter.cs
andrewslavin: Can I confirm what you mean? Are you saying that if I checked `_compressed` for the exact match and it's not there, there's no point checking `_notCompressed` for partial match, because the end result will be the same?

Consider the following setup: `_compressed = ["image/foo", "*/*"]`, `_notCompressed = ["image/*"]`. If I'm testing `image/bar` and check `_notCompressed` for partial match, `SouldCompress` will return `false`. But if I don't check `_notCompressed` for partial match, it'll return `true` because of the `*/*` entry in `_compressed`.

src/Microsoft.AspNetCore.ResponseCompression/MimeTypeFilter.cs
andrewslavin: `MediaTypeHeaderValue.TryParse` is less restrictive than what I'm trying to achieve. E.g. it accepts `*/subtype` or `type/sub*`. If I remove validation, entries like that will be just ignored, which can be an element of surprise. If you still think it's better to not validate, will be happy to remove.

src/Microsoft.AspNetCore.ResponseCompression/MimeTypeFilter.cs
Tratcher: Ah, I hadn't considered the `*/*` case.

andrewslavin:
@Tratcher, @glennc Thank you for the feedback. I left two specific questions to your comments above, can you please answer those.

Regarding "reorganizing to separate value storage from business logic", I presume the main reason is to provide ability to override the existing implementation, because currently it is completely locked down. The storage here has been optimised for the biz logic requirements, so instead of separating them, can I suggest the following:

1. Introduce interface `IMimeTypeFilter`.
1. Change `MimeTypeFilter` to `DefaultMimeTypeFilter : IMimeTypeFilter`, with internal implementation still combining the logic and the storage.

This way developers will be able to either inherit from `DefaultMimeTypeFilter` and override whatever they need while still using the current storage design, or they can write `MyMimeTypeFilter : IMimeTypeFilter` from scratch and store the data internally as they wish.

Will this design be acceptable?

andrewslavin:
@Tratcher Thank you for the feedback. I now intend to do the following:

1. Separate logic and storage.
1. Make logic overridable.
1. Make MIME type filter optional when configuring `ResponseCompressionOptions`, so if you do not configure it, you get the current behaviour.

With point 3 above, will this still be too breaking for 2.1 release? If not, how much time do I have?

Tratcher:
That would certainly be an improvement. I still need to get this cleared with our PM @glennc.

We're trying to wrap up preview2 next week, after that it will be more difficult to accept new API surface like this.

andrewslavin:
@Tratcher, @glennc I've made the discussed changes: separated logic from data, made logic overridable, made the filter an optional opt-in. Please have a look.

andrewslavin:
@Tratcher, @glennc Any update?

Tratcher:
Thanks for the reminder and the work you've put in so far. The current version still has some design issues we'll need to iterate on. We're busy wrapping up 2.1 and decided we don't have bandwidth to tackle this until 2.2. In the meantime I'll try to give some periodic feedback.

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionOptions.cs
```diff
@@ -15,9 +15,14 @@ public class ResponseCompressionOptions
         /// </summary>
         public IEnumerable<string> MimeTypes { get; set; }
 
+        /// <summary>
+        /// Configures which MIME types will be compressed.
+        /// </summary>
+        public IMimeTypeFilter MimeTypeFilter { get; set; }
```
Tratcher: Putting an interface onto Options is going to be an maintenance barrier. We won't be able to add to or modify the interface without breaking implementers. One way of dealing with this is to remove the interface and use a concrete type, making sure everything is virtual so it can be overridden.

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionOptionsExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+            );
+
+        /// <summary>
+        /// Use MIME type filter with the specified implementation.
+        /// </summary>
+        /// <param name="responseCompressionOptions">The <see cref="ResponseCompressionOptions"/> to set the MIME type filter for.</param>
+        /// <param name="configureFilter">MIME type filter configuration actions.</param>
+        /// <param name="mimeTypeFilter">MIME type filter implementation.</param>
+        /// <returns>The <see cref="ResponseCompressionOptions"/> so that additional calls can be chained.</returns>
+        public static ResponseCompressionOptions UseMimeTypeFilter(
```
Tratcher: I'm not sure why this extension exists. Why pass the type and an action that operates on the type you just passed me?

src/Microsoft.AspNetCore.ResponseCompression/DefaultMimeTypeFilterConfig.cs
```diff
@@ -0,0 +1,113 @@
...
+namespace Microsoft.AspNetCore.ResponseCompression
+{
+    /// <inheritdoc/>
+    public class DefaultMimeTypeFilterConfig : IMimeTypeFilterConfig
+    {
+        private readonly HashSet<string> _compressed = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+        private readonly HashSet<string> _notCompressed = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+
+        /// <inheritdoc/>
+        public IEnumerable<string> CompressedMimeTypes => _compressed;
```
Tratcher: If these sets were exposed as mutable collections then all of the methods could be removed from this type and become extension methods. They're all syntax sugar over the collections and I'm not sure we need most of them. The primary operation will be to add one or more entries to either list.

Tratcher: This collection is practically the same as Options.MimeTypes. We could consolidate the storage to the base Options class with two properties, the existing MimeTypes and a new ExcludedMimeTypes. Then IMimeTypeFilterConfig goes away entirely.

src/Microsoft.AspNetCore.ResponseCompression/DefaultMimeTypeFilter.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        /// <inheritdoc/>
+        public IMimeTypeFilterConfig Config { get; }
+
+        /// <inheritdoc/>
+        public DefaultMimeTypeFilter(IMimeTypeFilterConfig mimeTypeFilterConfig) =>
+            Config = mimeTypeFilterConfig ?? throw new ArgumentNullException(nameof(mimeTypeFilterConfig));
+
+        /// <inheritdoc/>
+        public virtual bool ShouldCompress(string mimeType) =>
```
Tratcher: Assuming the Config gets merged into Options as I suggested elsewhere, this ShouldCompress logic can move into ResponseCompressionProvider. That's already a replaceable component.

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionProvider.cs
```diff
@@ -113,21 +119,22 @@ public virtual bool ShouldCompressResponse(HttpContext context)
...
 
-            var separator = mimeType.IndexOf(';');
-            if (separator >= 0)
+            var separatorPos = mimeType?.IndexOf(';');
+
+            if (separatorPos >= 0)
             {
                 // Remove the content-type optional parameters
-                mimeType = mimeType.Substring(0, separator);
+                mimeType = mimeType.Substring(0, separatorPos.Value);
```
Tratcher: We should consider using StringSegment or Span here to reduce allocations.

src/Microsoft.AspNetCore.ResponseCompression/DefaultMimeTypeFilter.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        private bool? ShouldCompressPartial(string mimeType)
+        {
+            int? slashPos = mimeType?.IndexOf('/');
+
+            if (slashPos > -1)
+            {
+                string partialMimeType = $"{mimeType.Substring(0, slashPos.Value)}/*";
+                return ShouldCompressExact(partialMimeType);
```
Tratcher: We should write a version of ShouldCompressExact that doesn't require constructing a new partialMimeType. MediaTypeHeaderValue.IsSubsetOf does something like this:
https://github.com/aspnet/HttpAbstractions/blob/e80d0b000ac6b9a32ddaf01dc89b2c0b9a1d0523/src/Microsoft.Net.Http.Headers/MediaTypeHeaderValue.cs#L385

andrewslavin:
Thanks @Tratcher, all comments make sense and will make this PR much smaller and simpler.

I have one question - regarding StringSegments and MediaTypeHeaderValue.IsSubsetOf(). I do not see how I can use them, and ALSO preserve the current approach of looking up the hashsets using HashSet.Contains. To check if response MIME type is in the hashset, I will have to do something like this:

```
var separatorPos = mimeType.IndexOf(';');
var mimeTypeStringSegment = new StringSegment(mimeType).Subsegment(0, separatorPos).Trim();
var contains = hashset.Any(mt => mimeTypeStringSegment.Equals(mt, StringComparison.OrdinalIgnoreCase));
```

I did some simple performance testing. As expected, the speed of allocating a string and calling HashSet.Contains does not depend on the hashset size, while the above approach takes longer as the hashset size increases. Hashset size when both approaches take the same time is ~15 elements.

Do you still want me to proceed with StringSegments using the above implementation? Or is there a different way I'm missing?

Tratcher:
Don't worry about the string segment too much, if there's an easy way to use it then do so.

Tratcher:
Much better. Want to update the example to show wildcards as well as the new excludes?

andrewslavin:
@Tratcher I have made the requested changes, the whole thing is much simpler now.

On a different note, I was curious about StringSegments. I noticed that you replaced `string.Equals()` with `StringSegment.Equals()` in one of the PRs. The arguments to the function are two strings, so no new sting allocations happen in either case, while `StringSegment.Equals()` results in two StringSegment instances created out of two strings. I did some quick perf. testing, and though both functions are obviously very fast, `string.Equals()` is 3-4 times faster than `StringSegment.Equals()` in this case.

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionProvider.cs
```diff
@@ -49,14 +52,21 @@ public ResponseCompressionProvider(IServiceProvider services, IOptions<ResponseC
...
+            _excludedMimeTypes = new HashSet<string>(
+                responseCompressionOptions.ExcludedMimeTypes ?? Enumerable.Empty<string>(),
+                StringComparer.OrdinalIgnoreCase
+            );
+
+            _mimeTypes = new HashSet<string>(
+                responseCompressionOptions.MimeTypes ?? Enumerable.Empty<string>(),
+                StringComparer.OrdinalIgnoreCase
+            );
+            if (_mimeTypes.Count == 0 && _excludedMimeTypes.Count == 0)
```
Tratcher: Setting an exclusion wipes out the inclusion defaults? What's the point of exclusions if you don't have any inclusions?

You're missing a test case for this. E.g. No includes specified and one exclude specified. The default includes should still apply.

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionProvider.cs
```diff
@@ -126,8 +136,10 @@ public virtual bool ShouldCompressResponse(HttpContext context)
                 mimeType = mimeType.Trim();
             }
 
-            // TODO PERF: StringSegments?
-            return _mimeTypes.Contains(mimeType);
+            return ShouldCompressExact(mimeType) //check exact match type/subtype
+                ?? ShouldCompressPartial(mimeType) //check partial match type/*
+                ?? ShouldCompressExact("*/*") //check wildcard */*
```
Tratcher: Can exclude `*/*` ever override anything except an include `*/*`? Not sure why someone would define both. This could be simplified to only check for include `*/*`.

src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionProvider.cs
```diff
@@ -139,5 +151,34 @@ public bool CheckRequestAcceptsCompression(HttpContext context)
...
+            return null;
+        }
+
+        private bool? ShouldCompressPartial(string mimeType)
+        {
+            int? slashPos = mimeType?.IndexOf('/');
+
+            if (slashPos >= 0)
+            {
+                string partialMimeType = $"{mimeType.Substring(0, slashPos.Value)}/*";
```
Tratcher: string concat (+) is preferred over interpolation ($). It's simpler and faster.

Tratcher:
Where are we using StringSegment.Equals with strings? Can you give a link?

andrewslavin:
@Tratcher regarding StringSegment.Equals with strings - sorry, my mistake. I was talking about this PR: https://github.com/aspnet/BasicMiddleware/pull/235/files, but one of the two arguments there is actually a StringSegment.

andrewslavin:
Hi @Tratcher, I've made all the requested changes. Please have a look.

I would also like to provide some feedback about the implementation we ended up with. It tries to fix middleware configurations that we presume to be misconfigurations - namely uses default MIME times if no included MIME types were specified, and ignores excluded `*/*` if included `*/*` was specified. While I see the merit of this approach, as a middleware consumer I personally would prefer the "what you see is what you get" behaviour, i.e. what I specified when configuring options is what will get used. Changing the configured options behind the scenes is unexpected behaviour, and the end result is not necessarily what the consumer might have wanted. A less likely scenario is that the incl. and excl. MIME types come from an external source and/or get calculated using some logic, and the incl. list ends up empty. A more likely scenario is that I want to temporarily turn compression off to investigate some issue, and I do it via setting included MIME types to empty list or null.

I think having MimeTypeFilter separately from MimeTypes was actually a good design. It was providing an "I know what I'm doing and want full control" alternative to the "safe" MimeTypes list approach.

Tratcher:
Understood on the options fixup. I'd have to dig into the original discussions from when we designed it this way, there were some trade offs we were trying to ballance.

I do like the simplicity of the current design and the code looks good. I've asked @glennc to take a second look at the usability.

andrewslavin:
Yes, this is very simple and I like that too. If it was not for the two options overrides, I'd be very happy with this implementation. The `*/*` override is new and we can back out of it, but we cannot back out of the empty list override. That's the only reason I'd consider MimeTypeFilter. We can significantly simplify it, like you already mentioned in your comments.

andrewslavin:
Hi @Tratcher, any update on this PR?

Tratcher:
@andrewslavin sorry about the delay. I've rebased this on release/2.2, resolved the conflicts, and [merged](https://github.com/aspnet/BasicMiddleware/commit/cc610955dbc9aea4b649588779c49a4bfbf72630) it. Thanks for the new feature.

Issue #317, opened 2018-04-05
HostFiltering should honor realtime options updates (inc. via configuration)


DamianEdwards:
It would be great to be able to change the configured hosts for the app without having to restart it. This likely requires a change in `WebHost` too, around [here](https://github.com/aspnet/MetaPackages/blob/dev/src/Microsoft.AspNetCore/WebHost.cs#L187).

DamianEdwards:
@Tratcher @muratg

Tratcher:
Hmm, I don't think this is a feature we've supported in any middleware yet. We have reloadable config, but do we use it ourselves anywhere? Logging? @haok

DamianEdwards:
We use it in logging for sure, including the Azure App Service integration. In trying this feature is found myself wanting to change the value and have the app honor it.

HaoK:
It shouldn't be too hard to switch the middleware over, just need to hold onto IOptionsMonitor instead of caching the TOptions instance and use monitor.CurrentValue to access the options value.

```C#
        public HostFilteringMiddleware(RequestDelegate next, ILogger<HostFilteringMiddleware> logger, 
            IOptionsMonitor<HostFilteringOptions> options)
        {
            _next = next ?? throw new ArgumentNullException(nameof(next));
            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
            _options = options ?? throw new ArgumentNullException(nameof(options));
        }

        private HostFilteringOptions Options { get => _options.CurrentValue }
```

muratg:
@DamianEdwards Do you want to timebox this (say "Small") in 2.1?

DamianEdwards:
Yeah, it shouldn't be any harder than what @HaoK said plus ensuring the file config added by default flows changes (which I'm sure it already does).

PR #320, opened 2018-04-11
Reload HostFilter options on change

Tratcher:
#317 @DamianEdwards requested this magic.

As follow up, the code from the sample Startup will be used to update CreateDefaultBuilder:
https://github.com/aspnet/MetaPackages/blob/f972d15501fa494e39ffc81ef3a594b64780ca4a/src/Microsoft.AspNetCore/WebHost.cs#L187-L191

### DIFF
--- a/samples/HostFilteringSample/Startup.cs
+++ b/samples/HostFilteringSample/Startup.cs
@@ -1,12 +1,15 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.HostFiltering;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
 
 namespace HostFilteringSample
 {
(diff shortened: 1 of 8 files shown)


test/Microsoft.AspNetCore.HostFiltering.Tests/HostFilteringMiddlewareTests.cs
```diff
@@ -179,5 +183,58 @@ public class HostFilteringMiddlewareTests
...
+            var currentHost = "otherHost";
+
+            var builder = new WebHostBuilder()
+                .ConfigureServices(services =>
+                {
+                    services.AddHostFiltering(options =>
+                    {
+                        options.AllowedHosts = new[] { config["AllowedHosts"] };
+                    });
+                    services.AddSingleton<IOptionsChangeTokenSource<HostFilteringOptions>>(new ConfigurationChangeTokenSource<HostFilteringOptions>(config));
```
HaoK: This won't be needed once the other PR is in right?

HaoK: Or is this testing a different API path?

Tratcher: This mimics what happens in CreateDefaultBuilder, but we can't call that from this repo.

src/Microsoft.AspNetCore.HostFiltering/HostFilteringMiddleware.cs
```diff
@@ -72,19 +81,20 @@ public Task Invoke(HttpContext context)
             return _next(context);
         }
 
-        private void EnsureConfigured()
+        private IList<StringSegment> EnsureConfigured()
```
HaoK: Nit: This method is really GetAllowedHosts() now right?

samples/HostFilteringSample/Startup.cs
```diff
@@ -23,9 +26,22 @@ public void ConfigureServices(IServiceCollection services)
         {
             services.AddHostFiltering(options =>
             {
-                // If this is excluded then it will fall back to the server's addresses
-                options.AllowedHosts = Config.GetSection("AllowedHosts").Get<List<string>>();
+
```
HaoK: Is there a reason you prefer doing this in PostConfigure instead of in the Middleware, perhaps in the EnsureConfig method that computes the list instead?

That is generally safer than assuming no-one adds a PostConfigure after you

src/Microsoft.AspNetCore.HostFiltering/HostFilteringMiddleware.cs
HaoK: I think the PostConfigure logic should be moved into this method to ensure its run 'last'

samples/HostFilteringSample/Startup.cs
Tratcher: It's a compromise between Damian and David to avoid direct config lookups in the middleware. CreateDefaultBuilder has config layout opinions, not the individual components.

src/Microsoft.AspNetCore.HostFiltering/HostFilteringMiddleware.cs
Tratcher: It was David's decision to keep them seperate.

Issue #278, opened 2017-12-06
Add logging to HttpsPolicy middleware 


jkotalik:
Logging:
- When Hsts header is added or blocked
- HttpsRedirection, url redirect to.
- Any major error cases (misconfigured https port, etc.)

nbarbettini:
There is some logging now in HttpsRedirectionMiddleware, but not in HstsMiddleware. It would definitely be useful when trying to figure out why the header is or isn't being returned. I'll work up a PR.

PR #327, opened 2018-05-05
Add logging to HstsMiddleware

nbarbettini:
Added logging to HstsMiddleware in the style of HttpsRedirectionMiddleware.
- Debug log when HSTS is skipped because the request is insecure
- Debug log when HSTS is skipped because the host is excluded (by default or otherwise)
- Trace log when HSTS is added to the response.

Addresses part of #278


Tratcher:
Nicely executed. Thank you.

PR #349, opened 2018-07-11
Fixing universe

jkotalik:
Need to explicitly add the runtime framework version for 2.1

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -1,5 +1,6 @@
 <Project>
   <PropertyGroup>
+    <RuntimeFrameworkVersion Condition=" '$(TargetFramework)' == 'netcoreapp2.1' ">$(MicrosoftNETCoreApp21PackageVersion)</RuntimeFrameworkVersion>
     <RuntimeFrameworkVersion Condition=" '$(TargetFramework)' == 'netcoreapp2.2' ">$(MicrosoftNETCoreApp22PackageVersion)</RuntimeFrameworkVersion>
     <NETStandardImplicitPackageVersion Condition=" '$(TargetFramework)' == 'netstandard2.0' ">$(NETStandardLibrary20PackageVersion)</NETStandardImplicitPackageVersion>
     <!-- aspnet/BuildTools#662 Don't police what version of NetCoreApp we use -->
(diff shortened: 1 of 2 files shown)


Tratcher:
Your commit messages need work... 😁

Directory.Build.targets
```diff
@@ -1,5 +1,6 @@
 <Project>
   <PropertyGroup>
+    <RuntimeFrameworkVersion Condition=" '$(TargetFramework)' == 'netcoreapp2.1' ">$(MicrosoftNETCoreApp21PackageVersion)</RuntimeFrameworkVersion>
```
natemcmaster: What's still using netcoreapp2.1?

build/dependencies.props
```diff
@@ -24,6 +24,7 @@
     <MicrosoftExtensionsLoggingTestingPackageVersion>2.2.0-preview1-34640</MicrosoftExtensionsLoggingTestingPackageVersion>
     <MicrosoftExtensionsOptionsConfigurationExtensionsPackageVersion>2.2.0-preview1-34640</MicrosoftExtensionsOptionsConfigurationExtensionsPackageVersion>
     <MicrosoftExtensionsOptionsPackageVersion>2.2.0-preview1-34640</MicrosoftExtensionsOptionsPackageVersion>
+    <MicrosoftNETCoreApp21PackageVersion>2.1.0</MicrosoftNETCoreApp21PackageVersion>
```
natemcmaster: 2.1.2 is latest.

Directory.Build.targets
Tratcher: ResponseCompression cross compiles for netcoreapp2.1. Kestrel does something similar.

natemcmaster: Why? This is going into the 2.2 release, right? I think it should be lifted to target netcoreapp2.2.

jkotalik:
> Your commit messages need work... 😁

Sorry I have a script for feedback that I should adapt to take a commit message

Directory.Build.targets
Tratcher: Product packages are staying at minimum supported TFMs.

natemcmaster: There is no point in having this target netcoreapp2.1. We would not support mixing this package in with 2.1 dependencies. Because this package has overlapping dependencies with Microsoft.AspNetCore.App, it wouldn't work with Microsoft.AspNetCore.App anyways.

ryanbrandenburg:
Can we get Universe unblocked?

Tratcher:
@muratg has already said there's no plan to update product packages to 2.2. Let's resolve this offline.

jkotalik:
🆙 📅

Tratcher:
Resolved: We're sticking with lowest required TFMs for now and we expect everything to reset in 3.0.

Issue #358, opened 2018-09-07
IPv4 to IPv6 Mapped Addresses not mapped to IPv4 when checking known proxies/networks


SebastianC:
My scenario:
- web client goes through a load balancer
- load balancer directs to AWS/ECS host running a docker container
- docker container hosts a kestrel endpoint
- mvc controller in kestrel determines what the RemoteIPAddress is

My problem:
- When the docker container forwards the REST call to the Kestrel server, docker does an IPv4 to IPv6 conversion.
- When the Kestrel server attempts to evaluate the known proxies/networks, none of those evaluations handle the IPv4->IPv6 conversion so I always get that IP Address as the RemoteIPAddress (in my circumstance that's the converted form of the load balancer).

Proposed fix:
- In `Microsoft.AspNetCore.HttpOverrides.ForwardedHeadersMiddleware`  function `CheckKnownAddress`, check if `address.IsIPv4MappedToIPv6`, and if it is, map it back to the IPv4 address before continuing.  Pull request coming shortly.

This solution seems like it would also resolve another issue: #341 - "::ffff:127.0.0.1 not recognized as trusted local proxy"

Tratcher:
One trouble with this is that people have been adding the IPv6 versions as a workaround. Changing it internally will break them unless you also change the KnownNetworks values for them.

SebastianC:
I had originally considered updating another property to `ForwardedHeadersOptions` to indicate that this conversion should be turned "on", and default it to "off".  Would that resolve your concern?

If it's a KnownProxy, I can see how adding the mapped-value would work, but i was unable to create a KnownNetwork that would be matched appropriately.

Tratcher:
Checking both varians is a good mitigation.

How's this for a simplification?
```
        private bool CheckKnownAddress(IPAddress address)
        {
            if (address.IsIPv4MappedToIPv6)
            {
                var ipv4Address = address.MapToIPv4();
                if (CheckKnownAddress(ipv4Address))
                {
                   return true;
                }
            }

            if (_options.KnownProxies.Contains(address))
            {
                return true;
            }

            foreach (var network in _options.KnownNetworks)
            {
                if (network.Contains(address))
                {
                    return true;
                }
            }
            return false;
        }
```

SebastianC:
Better than mine. Will update the PR.

PR #359, opened 2018-09-07
Handle IPv4ToIPv6 Mapped Addresses when checking known proxies/networks

SebastianC:
Un-Map IPv4-to-IPv6-Mapped ips before comparing to known proxies/networks.

Addresses #358

### DIFF
--- a/src/Microsoft.AspNetCore.HttpOverrides/ForwardedHeadersMiddleware.cs
+++ b/src/Microsoft.AspNetCore.HttpOverrides/ForwardedHeadersMiddleware.cs
@@ -351,7 +351,11 @@ namespace Microsoft.AspNetCore.HttpOverrides
         {
             if (address.IsIPv4MappedToIPv6)
             {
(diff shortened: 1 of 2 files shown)


Tratcher:
As discussed on the issue, this needs to check both versions of the IP.

Tratcher:
The code looks good. Can you add some tests?

SebastianC:
I would love to, but running into the following issue when I go to run unit tests:

```
Testhost process exited with error: It was not possible to find any compatible framework version
The specified framework 'Microsoft.NETCore.App', version '2.2.0-preview1-26618-02' was not found.
```

Downloading the current preview version from https://www.microsoft.com/net/download/dotnet-core/2.2 gives me a different version:
```
> dotnet --version
2.2.100-preview1-009349
```

Do you know how I can go about resolving this?
@Tratcher

Tratcher:
See https://github.com/aspnet/Home/wiki/Building-from-source

PR #363, opened 2018-09-15
replacing string interpolation with proper logging syntax

### LINKED ISSUE #362: HostFilteringMiddleware uses C# interopolated strings for tracing
`HostFilteringMiddleware` is using C# interopolated strings instead of passing the arguments off to the `ILogger`. I'm using Serilog which will cache the trace message, so if that changes, it will balloon. Plus, its causing more work for the middleware when the tracing is not turned up.

daniel-white:
Fixes #362.

* I found some other places where it was used.
* I'm not sure of the convention of the logging properties
* I couldn't get the preview bits to build on my Mac or AppVeyor.
* I'm not sure if the tests will pass


daniel-white:
well your CI got it passed. awesome sauce.

Tratcher:
Thanks

Issue #222, opened 2017-04-03
Add logging to the response compression middleware


Tratcher:
Review the logs for the following scenario:
https://github.com/aspnet/StaticFiles/issues/7#issuecomment-291021369

Compression was happening, but it never showed up in the logs.

Things to log:
- Compressing response with encoding x
- Not compressing response due to: no matching content-type, no matching accept-encoding values, etc..

PR #371, opened 2018-09-28
Add logging to the response compression middleware

Tratcher:
#222 No functional changes, only minor re-organization to facilitate logging. Recommend reviewing with Hide whitespace changes enabled. Also consolidated some of the test logic.

### DIFF
--- a/src/Microsoft.AspNetCore.ResponseCompression/internal/ResponseCompressionLoggingExtensions.cs
+++ b/src/Microsoft.AspNetCore.ResponseCompression/internal/ResponseCompressionLoggingExtensions.cs
@@ -1,0 +1,72 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Logging;
+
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.ResponseCompression/internal/ResponseCompressionLoggingExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            _noCompressionForContentType(logger, header, null);
+        }
+
+        public static void ShouldCompressResponse(this ILogger logger)
+        {
+            _shouldCompressResponse(logger, null);
+        }
+
+        public static void NoCompressionNoProvider(this ILogger logger)
```
JunTaoLuo: Why the duplicate `No`?

JunTaoLuo: Oh never mind. I see the pattern.

JunTaoLuo: Reads a little awkwardly though, I think `NoCompressionProvider` and `NoAcceptEncoding` would be better.

PR #375, opened 2018-10-17
Specify culture to int.TryParse

Tratcher:
There's guidance to always specify the culture when parsing numbers, otherwise the data may be misinterpreted on some machines depending on the local culture. #356 called this to our attention and these changes should mitigate it. Note there's a deeper corefx issue there that I do not have a repro for and I'll move that to corefx after merging this.


Tratcher:
I tried adding that test but was not able to repro the original users issue. And you're right, messing with the culture is problematic.

Issue #376, opened 2018-10-23
Enable response compression support for WebAssembly by default


danroth27:
Currently for Blazor we explicitly enable response compression support for WebAssembly (`application/wasm`). Given that WebAssembly is picking up, it makes sense to enable response compression for WebAssembly by default.

Tratcher:
@muratg @shirhatti this is a XS ask for 2.2.

muratg:
@Tratcher what's the PR looking like? Single line addition to https://github.com/aspnet/BasicMiddleware/blob/release/2.2/src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionDefaults.cs? 

I think it's fine, but I'll let @shirhatti make a decision.

Tratcher:
Right. I'll send it.

PR #377, opened 2018-10-23
Add application/wasm to the default compression list

Tratcher:
@danroth27 for use with blazor and other web assembly frameworks #376


src/Microsoft.AspNetCore.ResponseCompression/ResponseCompressionDefaults.cs
```diff
@@ -27,6 +27,8 @@ public class ResponseCompressionDefaults
             "text/xml",
             "application/json",
             "text/json",
+            // Other
```
shirhatti: Nit: I'd rather say WebAssembly. Not a fan of catch-all terms

